PR #7, opened 2026-06-11
chore: scaffold pnpm monorepo with MF host shell and Go server

kellylimmm:
Closes https://github.com/String-sg/teacher-workspace/issues/1

Sets up the foundational repository structure for teacher-workspace so feature work can begin against an agreed layout without each developer configuring tooling from scratch.

## Changes
- **pnpm workspace** — `pnpm-workspace.yaml` scans `apps/*`; all front-end apps
  are scoped `@teacher-workspace/<name>`
- **MF host shell** (`apps/host`) — Rsbuild + `@module-federation/rsbuild-plugin`,
  configured as a Module Federation host with no remotes wired yet; React 19 and
  react-dom declared as eager shared singletons
- **Go server** (`server/main.go`) — bare entry point under module path
  `github.com/String-sg/teacher-workspace`; routing and handlers are out of scope
  for this story
- **Toolchain pinning** — `engines` in `package.json` pins Node `^24.0.0` and
  pnpm `^11.0.0`; `engine-strict=true` in `.npmrc` makes mismatches a hard
  failure; `go.mod` pins Go 1.26 with `toolchain go1.26.0`
- **`.gitignore`** — matches the spec exactly, including `.vscode/*` with
  allow-list exceptions for `extensions.json` and `settings.json`

## Out of scope
- MF remotes (live in separate repositories)
- Shared packages beyond the host shell
- Business logic in the host shell or Go server
- CI/CD, lint/format config, git hooks
- Deployment and environment config

## Notes
- The async bootstrap pattern (`index.ts` → `bootstrap.tsx`) is required even
  with MF 2.0 — it defers React imports until after the MF runtime has
  initialised the shared scope. Removing it causes RUNTIME-006.
- `eager: true` on shared React is set on the host because there are no remotes
  yet; without it the shared scope has no factory to resolve against and throws
  at runtime.


server/main.go
```diff
@@ -0,0 +1,3 @@
+package main
+
+func main() {}
```
YimingIsCOLD: At least print something so we can test if it is working.

.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
YimingIsCOLD: Have you tested if this will work?

package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "teacher-workspace",
+  "version": "0.0.1",
+  "private": true,
+  "packageManager": "pnpm@11.0.0",
```
YimingIsCOLD: Do we still need this if there is already a minimum version in `engines` field?

package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "teacher-workspace",
+  "version": "0.0.1",
```
YimingIsCOLD: Should use `0.0.0` as we do not have any release yet.

apps/host/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "@teacher-workspace/host",
+  "version": "0.0.1",
```
YimingIsCOLD: Should use `0.0.0` as we do not have any release yet.

apps/host/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "dev": "rsbuild dev",
+    "build": "rsbuild build",
+    "preview": "rsbuild preview"
+  },
+  "dependencies": {
+    "react": "^19.0.0",
+    "react-dom": "^19.0.0"
+  },
+  "devDependencies": {
+    "@module-federation/rsbuild-plugin": "^0.9.0",
```
YimingIsCOLD: Please check the dependencies version. Some of it are outdated. Not sure if it will cause any problem if we only pin the major version, usually it should reflect the latest version e.g. `react: "^19.2.7"` instead of `react: "^19.0.0"`.

apps/host/rsbuild.config.ts
```diff
@@ -0,0 +1,28 @@
...
+export default defineConfig({
+  plugins: [
+    pluginReact(),
+    pluginModuleFederation({
+      name: 'host',
+      remotes: {},
+      shared: {
+        react: {
+          singleton: true,
+          eager: true,
```
YimingIsCOLD, line 14: Is it because we want `react` and `react-dom` to immediately load the shared module therefore we set `eager` to `true`?

apps/host/rsbuild.config.ts
```diff
@@ -0,0 +1,28 @@
+import { defineConfig } from '@rsbuild/core';
+import { pluginReact } from '@rsbuild/plugin-react';
+import { pluginModuleFederation } from '@module-federation/rsbuild-plugin';
+
+export default defineConfig({
+  plugins: [
+    pluginReact(),
+    pluginModuleFederation({
+      name: 'host',
```
YimingIsCOLD: For remotes, I will assume their names to be `student_insights` or `parent_gateway`, do we want something related to teacher workspace?

apps/host/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
YimingIsCOLD, line 1: I would recommend placing this file in either `apps/host/src/index.html` or `apps/host/index.html`. Files inside public folder is usually for static assets that can be served directly without modifications but I would treat `index.html` as application code because it will be used as a template for the bundler to inject additional code.

apps/host/tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "bundler",
+    "jsx": "react-jsx",
+    "strict": true,
+    "skipLibCheck": true,
+    "lib": ["ES2022", "DOM", "DOM.Iterable"]
+  },
+  "include": ["src", "rsbuild.config.ts"]
+}
```
YimingIsCOLD:
```suggestion
{
  "compilerOptions": {
    "lib": ["DOM", "DOM.Iterable", "ES2023"],
    "jsx": "react-jsx",
    "target": "ES2023",
    "noEmit": true,
    "skipLibCheck": true,

    /* modules */
    "module": "ESNext",
    "isolatedModules": true,
    "resolveJsonModule": true,
    "moduleResolution": "Bundler",
    "allowImportingTsExtensions": true,

    /* type checking */
    "strict": true,
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "paths": {
      "*": ["./@mf-types/*"],
      "~/*": ["./src/*"]
    }
  },
  "include": ["src", "rsbuild.config.ts"]
}
```

This is what I have gathered during the spike.

.gitignore
```diff
@@ -0,0 +1,26 @@
...
+
+# Ignore node modules.
+node_modules
+
+# Ignore cache/output directories.
+bin
+build
+dist
+out
+tmp
```
YimingIsCOLD, line 22: Might need to add `@mf-types` after you update the dependencies to the latest rsbuild core and plugins.

.npmrc
kellylimmm: Yeah but i found out that `engine-strict=true` applies to the packages pnpm installs from the registry, it does not apply to the root package. So it only warns about the root's engines mismatch instead of failing.

I have added a `preinstall check-engines` script so that `pnpm install` fails with a clear error on wrong Node/pnpm.

package.json
kellylimmm: `packageManager` tells tooling which package manager to use and enables corepack enforcement. `engines` declares the version range that works. They're complementary and doesn't replace the other.

apps/host/rsbuild.config.ts
kellylimmm: `eager: true` is a workaround for having no remotes yet. Without remotes, MF has nobody to negotiate a shared React version with, so the negotiation stalls and throws an error. Setting `eager: true` skips the negotiation entirely and bundles React directly into the main chunk. Once remotes are added, we can revisit this.

kellylimmm: Another workaround is to remove shared from the config entirely for now.

apps/host/index.html
kellylimmm: I chose `apps/host/index.html` as it makes it clear that it's the app entry point, separate from the source modules.

pnpm-workspace.yaml
nicholasjjlim, line 1: suggest to add `minimumReleaseAge: 10080` (7d) here to reduce supply chain compromise when installing or updating dependencies

https://pnpm.io/settings#minimumreleaseage

package.json
YimingIsCOLD: We are not using corepack so no need `packageManager`?

.npmrc
YimingIsCOLD: I think pnpm can already check engine. Just need to place the settings in the correct place. We don't need `.npmrc` for now because we are not publishing or specifying new registry.

https://pnpm.io/settings#enginestrict

apps/host/rsbuild.config.ts
YimingIsCOLD: We can revist again once remotes are added.

.npmrc
kellylimmm: Do we still want to include the `preinstall check-engines scripts`?

pnpm-workspace.yaml
kellylimmm: Curious why `10080` and not other amount? 

Also, not sure if it's part of the scope of this story since i have no Renovate set up yet.

.npmrc
YimingIsCOLD: I think no need as pnpm already check engines.

pnpm-workspace.yaml
kellylimmm: I added minimumReleaseAge to this pr based on our convo offline. 

> Curious why `10080` and not other amount?

Standard days used by GT as mentioned by Nic.

> Also, not sure if it's part of the scope of this story since i have no Renovate set up yet.

No harm setting up early.

.npmrc
kellylimmm: Removed preinstall scripts

YimingIsCOLD (approved):

PR #29, opened 2026-06-24
feat(`server/middleware`): add request ID and request logging middleware

### LINKED ISSUE #9: feat(`server/middleware`): add request ID and request logging middleware
## User story

As a developer operating the teacher-workspace server, I want every HTTP request to carry a unique request ID and emit a structured access-log line, so that I can trace a request across all its log output and observe its method, path, status, and latency.

## Background

Follow-up to #6, which scaffolds the Go HTTP server (slog default logger, `http.NewServeMux`) and defers request-logging middleware to a follow-up.

This issue adds two `func(http.Handler) http.Handler` middleware:

1. **Request ID**: generates a unique ID per request, sets it as the `X-Request-ID` response header, derives a request-scoped slog logger carrying the ID, and stores that logger on the request context. The ID algorithm is ported from `teacher-workspace-archive`, not reinvented.
2. **Request log**: reads the request-scoped logger from the context and, after the handler returns, emits one structured line with method, path, status, and duration.

Because handlers read the logger from the context, every line produced while serving a request is tagged with the same request ID, making it traceable end to end.

### Request is logged with structured fields

- **Given** the server is running with both middleware installed
- **When** a client sends a request to any route
- **Then** a single structured log line is emitted recording the method, path, response status code, and duration

### Each request receives a unique request ID

- **Given** the server is running with both middleware installed
- **When** a client sends two separate requests
- **Then** each request's log line carries a request ID attribute, and the two IDs differ

### Response carries the request ID header

- **Given** the server is running with both middleware installed
- **When** a client sends a request to any route
- **Then** the response includes an `X-Request-ID` header whose value matches the request ID in that request's access-log line

### Request ID propagates to handler log lines

- **Given** a handler that logs using the logger from the request context
- **When** a client sends a request to that handler
- **Then** the handler's log lines and the access-log line all carry the same request ID

### Response without an explicit status is logged as 200

- **Given** a handler that writes a body without setting a status code
- **When** a client sends a request to that handler
- **Then** the access-log line records the status as 200

### Not found responses are logged

- **Given** the server is running with both middleware installed
- **When** a client sends a request to an unregistered path
- **Then** an access-log line is still emitted recording the path and status 404

## Out of scope

- Honouring an inbound `X-Request-ID` header; the ID is always generated server-side.
- Distributed tracing, span/trace correlation, or OpenTelemetry.
- Other middleware (auth, CORS, panic/recovery, compression).
- Making log fields, format, or sampling configurable; middleware use the slog default logger from #6.
- Logging request or response bodies or headers.

## Design assets

No Figma assets. Per-request middleware flow:

```mermaid
sequenceDiagram
  participant client as Client
  participant idmw as RequestID middleware
  participant logmw as RequestLog middleware
  participant h as Handler

  client->>idmw: HTTP request
  idmw->>idmw: generate request ID (ported algo)
  idmw->>idmw: set X-Request-ID response header
  idmw->>idmw: logger = slog.Default().With(request_id)
  idmw->>logmw: request with logger in context
  logmw->>logmw: start timer, wrap ResponseWriter to capture status
  logmw->>h: serve request
  h->>h: logs via logger from context (tagged with request_id)
  h-->>logmw: response written
  logmw->>logmw: log method, path, status, duration via context logger
  logmw-->>client: response
```

---

## Technical context

- Server code lives under `server/` per the #6 layout: `server/cmd/tw` (entrypoint that configures the slog default logger and builds the mux), `server/internal` (e.g. `server/internal/handler`).
- Routing uses `http.NewServeMux`; middleware are `func(http.Handler) http.Handler` wrappers applied around the mux.
- The request ID algorithm exists in `teacher-workspace-archive` and should be ported from there.

### Package and file placement

- New package `server/internal/middleware`.
- `requestid.go`: request ID middleware, ID generation function, unexported context key, exported logger accessor.
- `requestlog.go`: request log middleware and its `http.ResponseWriter` wrapper.

### Request ID generation

- Port the algorithm from `teacher-workspace-archive`, keeping it behaviourally equivalent. Generation has no error path that blocks the request; it produces an ID for every request.

### Request ID middleware

- Set the ID via `w.Header().Set("X-Request-ID", id)` before calling next.
- Build the logger as `slog.Default().With(<attr>, <id>)` and store it on the context under an unexported key type (not a bare string), then call next with `r.WithContext(ctx)`.
- Use one attribute key (e.g. `request_id`) consistently.

### Logger accessor

- `func LoggerFromContext(ctx context.Context) *slog.Logger` returns the request logger when present, falling back to `slog.Default()` so callers never get nil.

### Status and duration capture

- Record start time on entry; compute duration with `time.Since` after next returns.
- Wrap the `http.ResponseWriter` in a recorder that captures the `WriteHeader` status, defaults to 200 (so body-only handlers log as 200), and records only the first `WriteHeader` call.
- The recorder implements `Unwrap() http.ResponseWriter` so `http.NewResponseController` can reach the underlying writer (Flush, deadlines).
- Emit the log line via the context logger (carrying the request ID) with method, path, status, and duration.

### Ordering and wiring

- In `server/cmd/tw`, wrap the mux with request ID on the outside and request log on the inside, so the ID and logger (and `X-Request-ID` header) are set before the log middleware runs. Reversed, the log line would fall back to the default logger and omit the ID.

### Log middleware tolerates a missing request logger

- **Given** the request log middleware runs without request ID middleware ahead of it
- **When** a request passes through
- **Then** it logs via the default logger without panicking, still recording method, path, status, and duration

### Concurrent requests get independent request IDs

- **Given** the server handles multiple requests concurrently
- **When** they are processed
- **Then** each has its own request ID and request-scoped logger with no shared mutable state, and the race detector reports no data race

### Status recorder captures the first WriteHeader only

- **Given** a handler that calls `WriteHeader` with a non-200 status then returns
- **When** the log middleware records the status
- **Then** the logged status matches the first `WriteHeader` value

### ResponseController reaches the underlying writer

- **Given** a handler that obtains an `http.ResponseController` and calls `Flush`
- **When** the request is served through the log middleware
- **Then** the call resolves to the underlying writer via `Unwrap` without error

## Hard constraints

- No third-party dependency for ID generation or middleware; port from `teacher-workspace-archive` and use only the standard library (`net/http`, `log/slog`, `context`, `time`).
- No router or middleware framework; use `func(http.Handler) http.Handler` and `http.NewServeMux` from #6.
- Do not store the request ID under a bare `string` context key; use an unexported key type.
- Do not log request or response bodies or headers.## User story

---

> 🌊 Managed by **[Currents](https://tfx-currents.up.railway.app)** · [Task 9320a754-40e1-49df-b316-b9084bd9d406](https://tfx-currents.up.railway.app/?task=9320a754-40e1-49df-b316-b9084bd9d406)

{"metadata":{"sourceType":"github"},"rank":"Zz","taskId":"9320a754-40e1-49df-b316-b9084bd9d406","version":1}

kellylimmm:
Closes #9

## Summary

- Adds `RequestID` middleware that generates a unique 32-char base58 ID per request, sets the `X-Request-ID` response header, and stores a request-scoped slog logger (tagged with `request_id`) on the context
- Adds `RequestLog` middleware that wraps the `ResponseWriter` to capture status (defaults to 200) and emits one structured access-log line after the handler returns with method, path, status, and duration
- Wires both middleware in `server/cmd/tw` with `RequestID` on the outside and `RequestLog` on the inside so the ID and logger are in context before the log middleware runs
- Ports `pkg/random` (base58 ID generation algorithm) and `pkg/require` (test assertion helpers) from `teacher-workspace-archive`

## Changes

- `server/pkg/random/random.go` + `random_test.go` — ported cryptographically secure alphanumeric string generator
- `server/pkg/require/require.go` — ported test assertion helpers
- `server/internal/middleware/requestid.go` — `RequestID` middleware, unexported context key types, `LoggerFromContext` accessor
- `server/internal/middleware/requestid_test.go` — tests for all RequestID acceptance criteria
- `server/internal/middleware/requestlog.go` — `RequestLog` middleware, `responseRecorder` with `Unwrap()` for `ResponseController` compatibility
- `server/internal/middleware/requestlog_test.go` — tests for all RequestLog acceptance criteria and additional test scenarios
- `server/cmd/tw/main.go` — wired `RequestID(RequestLog(mux))` around the mux

## Test plan

- [x] Request is logged with structured fields (method, path, status, duration)
- [x] Two requests get different IDs
- [x] `X-Request-ID` header matches the context request ID
- [x] Handler logger is request-scoped (same `request_id` propagates)
- [x] No explicit `WriteHeader` logs status 200
- [x] Unregistered path logs status 404
- [x] Log middleware without `RequestID` ahead falls back to default logger without panicking
- [x] 50 concurrent requests with race detector reports no data race
- [x] First `WriteHeader` call is the one recorded
- [x] `ResponseController.Flush` resolves via `Unwrap` without error

### DIFF
--- a/server/cmd/tw/main.go
+++ b/server/cmd/tw/main.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/String-sg/teacher-workspace/server/internal/handler"
+	"github.com/String-sg/teacher-workspace/server/internal/middleware"
 )
 
 const (
@@ -21,9 +22,10 @@ const (
 func main() {
 	slog.SetDefault(slog.New(slog.NewJSONHandler(os.Stdout, nil)))
 
+	mux := handler.NewMux()
 	srv := &http.Server{
 		Addr:    addr,
-		Handler: handler.NewMux(),
+		Handler: middleware.RequestID(middleware.RequestLog(mux)),
 	}
 
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
--- a/server/internal/middleware/requestid.go
+++ b/server/internal/middleware/requestid.go
@@ -1,0 +1,48 @@
+package middleware
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+
+	"github.com/String-sg/teacher-workspace/server/pkg/random"
+)
+
+type ctxKeyRequestID struct{}
+type ctxKeyLogger struct{}
+
+const requestIDHeader = "X-Request-ID"
+
+// RequestID is an HTTP middleware that generates a unique request identifier
+// for each incoming request. The request ID and a request-scoped logger
+// containing it are stored in the request context. The ID is also written to
+// the response headers to support log correlation and request tracing.
+func RequestID(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		id := random.Base58(32)
+		ctx := context.WithValue(r.Context(), ctxKeyRequestID{}, id)
+
+		logger := slog.Default().With("request_id", id)
+		ctx = context.WithValue(ctx, ctxKeyLogger{}, logger)
+
+		w.Header().Set(requestIDHeader, id)
+
+		next.ServeHTTP(w, r.WithContext(ctx))
+	})
+}
+
+// RequestIDFromContext retrieves the request ID from the provided context.
+// The returned boolean indicates whether a request ID was present.
+func RequestIDFromContext(ctx context.Context) (string, bool) {
+	id, ok := ctx.Value(ctxKeyRequestID{}).(string)
+	return id, ok
+}
+
+// LoggerFromContext retrieves the request-scoped logger from the provided
+// context. If no logger is present, it falls back to the default logger.
+func LoggerFromContext(ctx context.Context) *slog.Logger {
+	if logger, ok := ctx.Value(ctxKeyLogger{}).(*slog.Logger); ok {
+		return logger
+	}
+	return slog.Default()
+}
--- a/server/internal/middleware/requestid_test.go
+++ b/server/internal/middleware/requestid_test.go
@@ -1,0 +1,116 @@
+package middleware
+
+import (
+	"context"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/String-sg/teacher-workspace/server/pkg/require"
+)
+
+func TestRequestID(t *testing.T) {
+	t.Run("sets X-Request-ID header on the response", func(t *testing.T) {
+		next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(http.StatusOK)
+		})
+
+		req := httptest.NewRequest(http.MethodGet, "/", nil)
+		rec := httptest.NewRecorder()
+
+		RequestID(next).ServeHTTP(rec, req)
+
+		res := rec.Result()
+		require.Equal(t, http.StatusOK, res.StatusCode)
+		require.Equal(t, 32, len(res.Header.Get(requestIDHeader)))
+	})
+
+	t.Run("two requests get different IDs", func(t *testing.T) {
+		var ids [2]string
+
+		next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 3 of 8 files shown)


yiningsoong:
LGTM 👍

---
*🤖 aif-code-review · bedrock.claude-opus-4-6[1m]*

yiningsoong (approved):

PR #38, opened 2026-06-30
feat(`server/config`): load configuration from environment variables and dotenv files

### LINKED ISSUE #8: feat(`server/config`): load configuration from environment variables and dotenv files
## User story

As a developer deploying the teacher-workspace server, I want configuration loaded from environment variables and an optional `.env` file with a clear precedence order, so that I can deploy with environment-specific settings instead of hardcoded values.

## Background

Issue #6 scaffolds the Go HTTP server but hardcodes its listen address (`:3000`), slog level, and `http.Server` timeouts, and defers "environment-based configuration" to this issue.

This adds a typed config layer resolving values in increasing precedence: built-in defaults < optional `.env` file < real process environment variables. All variables use the `TW_` prefix. The result is consumed by `server/cmd/tw/main.go`, replacing #6's hardcoded values. Loading is fail-fast: any unparseable or out-of-range value aborts startup with a slog error and a non-zero exit.

### Server runs on built-in defaults when nothing is configured

- **Given** no `TW_` variables are set and no `.env` file is present
- **When** the server is started
- **Then** config loads using defaults and the server listens on port 3000

### Real environment variable takes precedence over the .env file

- **Given** a `.env` file sets `TW_SERVER_PORT=8080` and the process env sets `TW_SERVER_PORT=9090`
- **When** the server is started
- **Then** the server listens on port 9090

### Value from the .env file is applied when the variable is unset

- **Given** a `.env` file sets `TW_LOG_LEVEL=debug` and `TW_LOG_LEVEL` is unset in the process env
- **When** the server is started
- **Then** the slog level is debug

### Missing .env file is tolerated

- **Given** no `.env` file exists
- **When** the server is started
- **Then** config loads without error using defaults and process env

### Invalid configuration aborts startup

- **Given** `TW_SERVER_PORT` is set to `"abc"` (or `TW_ENV` to an unknown value like `"staging"`)
- **When** the server is started
- **Then** loading fails with a clear slog error, the process exits non-zero, and the server never accepts connections

## Out of scope

- Config sources other than process env and a single optional `.env` file (no YAML/JSON/TOML, no CLI flags).
- Secrets management / vault integration.
- Hot-reloading; config is read once at startup.
- Committing a real `.env` (only a `.env.example` template is added; `.env` is git-ignored).
- TLS/HTTPS, CORS, and auth config (separate issues).
- Frontend/host config; this covers the Go server only.

## Design assets

No Figma assets. Configuration resolution order:

```mermaid
flowchart TD
  A[Built-in defaults] --> B{.env file present?}
  B -- yes --> C[Load .env: set only keys not already in process env]
  B -- no --> D[Skip .env without error]
  C --> E[Read effective values from process environment]
  D --> E
  E --> F{All values parse and validate?}
  F -- yes --> G[Return populated Config]
  F -- no --> H[Return error: main logs via slog and exits non-zero]
```

---

## Technical context

- Builds on #6: Go layout under `server/` (`server/cmd/tw` entrypoint, `server/internal` for modules), `go.mod` at repo root (module `github.com/String-sg/teacher-workspace`).
- Precedence (lowest to highest): defaults, optional `.env`, real process env. The dotenv semantic (a `.env` value applies only when the key is unset in the process env) yields exactly this ordering.

### Package and file placement

- `server/internal/config/config.go` (`package config`): `Config` type plus `Load() (Config, error)` that resolves defaults, optionally loads `.env`, reads process env, decodes/validates. Internal path may be adjusted provided `Load()` is consumed by `server/cmd/tw/main.go`.
- `server/internal/config/config_test.go`: table-driven tests using `t.Setenv` and temp `.env` files.
- `server/cmd/tw/main.go` (modify): call `config.Load()` first; on error log via slog and exit non-zero. Use the result for listen address, slog level, and `http.Server` timeouts.
- `.env.example` (repo root): documented template of every `TW_` variable with default and comment; no secrets.
- `.gitignore` (modify): add `.env` and `.env.*` except `.env.example`.

### Config schema and defaults

```go
type Environment string

const (
    EnvDevelopment Environment = "development"
    EnvProduction  Environment = "production"
)

type Config struct {
    Env      Environment // TW_ENV
    LogLevel slog.Level  // TW_LOG_LEVEL
    Server   ServerConfig
}

type ServerConfig struct {
    Port              int           // TW_SERVER_PORT
    ReadHeaderTimeout time.Duration // TW_SERVER_READ_HEADER_TIMEOUT
    ReadTimeout       time.Duration // TW_SERVER_READ_TIMEOUT
    WriteTimeout      time.Duration // TW_SERVER_WRITE_TIMEOUT
    IdleTimeout       time.Duration // TW_SERVER_IDLE_TIMEOUT
}
```

| Variable | Default | Parse / validation rule |
| --- | --- | --- |
| `TW_ENV` | `development` | Must be `development` or `production` |
| `TW_LOG_LEVEL` | `info` | One of `debug`/`info`/`warn`/`error` (case-insensitive) |
| `TW_SERVER_PORT` | `3000` | `strconv.Atoi`; 1-65535 |
| `TW_SERVER_READ_HEADER_TIMEOUT` | `2s` | `time.ParseDuration`; >= 0 |
| `TW_SERVER_READ_TIMEOUT` | `15s` | `time.ParseDuration`; >= 0 |
| `TW_SERVER_WRITE_TIMEOUT` | `30s` | `time.ParseDuration`; >= 0 |
| `TW_SERVER_IDLE_TIMEOUT` | `60s` | `time.ParseDuration`; >= 0 |

`TW_SERVER_PORT` must default to `3000` to preserve #6's behaviour when unconfigured.

### Precedence, .env loading, and validation

- Apply defaults, then `.env` values only for keys unset in the process env, then read effective values from the process env. A `.env` value must never override a real env var.
- `.env` parsing uses our hand-rolled dotenv library (ported from `teacher-workspace-archive`), not a third-party dotenv dependency.
- Each typed config carries its own `Validate() error` so rules live with the type (`Config.Validate` delegates to `ServerConfig.Validate`, `Environment.Validate`). `Load()` calls `Validate()` before returning, so a successful `Load` always yields a valid `Config`.
- **Open grooming decision**: how env values are decoded/bound into the typed `Config`: [`go-viper/mapstructure`](https://github.com/go-viper/mapstructure) plus glue vs a full framework like [`spf13/viper`](https://github.com/spf13/viper). Decide before implementation.

### .env populates only unset keys

- **Given** `.env` sets `TW_SERVER_PORT=8080` and `TW_LOG_LEVEL=debug`, and the process env sets `TW_SERVER_PORT=9090`
- **When** config is loaded
- **Then** Port is 9090 (process env wins) and LogLevel is debug (from `.env`)

### Duration and case-insensitive parsing

- **Given** `TW_SERVER_IDLE_TIMEOUT=90s` and `TW_LOG_LEVEL=DEBUG`
- **When** config is loaded
- **Then** `IdleTimeout` is 90s and `LogLevel` resolves to slog debug

### Out-of-range port is rejected by Validate

- **Given** `TW_SERVER_PORT=70000`
- **When** config is loaded
- **Then** `Validate` returns an error and no Config is used

## Hard constraints

- All variables use the `TW_` prefix and the exact names above.
- Precedence must be defaults < `.env` < real process env; a `.env` value must never override a real env var.
- The `.env` file must be optional (absence is not an error).
- `.env` parsing must use our hand-rolled dotenv library (from `teacher-workspace-archive`), not a third-party package.
- Validation rules must live in a `Validate() error` method on each typed config.
- Invalid/unparseable/out-of-range values must fail fast (return an error, exit non-zero); never fall back to a default for an explicitly provided value.
- `TW_SERVER_PORT` must default to `3000`.
- A real `.env` must not be committed; add `.env.example` and git-ignore `.env`.## User story

---

> 🌊 Managed by **[Currents](https://tfx-currents-prod.up.railway.app)** · [Task 7cd11da4-4a88-424a-acba-8ef36c8d6604](https://tfx-currents-prod.up.railway.app/?task=7cd11da4-4a88-424a-acba-8ef36c8d6604)

{"metadata":{"sourceType":"github"},"rank":"Zz","taskId":"7cd11da4-4a88-424a-acba-8ef36c8d6604","version":1}

kellylimmm:
Closes #8

## Summary

- Adds a typed `Config` struct loaded from built-in defaults, an optional `.env` file, and real process env variables (lowest to highest precedence)
- Ports the hand-rolled dotenv parser and loader from `teacher-workspace-archive` into `server/pkg/dotenv`
- Wires `config.Load()` into `main.go`, replacing all hardcoded values for port, log level, and `http.Server` timeouts

## Changes

| File | Change |
|---|---|
| `server/pkg/dotenv/parser.go` | Hand-rolled `.env` file parser, ported from archive |
| `server/pkg/dotenv/dotenv.go` | Loader: merges `.env` into process env for unset keys, decodes via mapstructure |
| `server/pkg/require/require.go` | Added `NoError` and `HasError` test helpers |
| `server/internal/config/config.go` | `Config`, `ServerConfig`, `defaults()`, `Load()`, `Validate()` |
| `server/internal/config/config_test.go` | Table-driven tests using `t.Setenv` and temp `.env` files |
| `server/cmd/tw/main.go` | Calls `config.Load()` first; uses result for slog level, addr, and all timeouts |
| `.env.example` | Documented template of all `TW_` variables |
| `.gitignore` | Updated to precise `.env` / `.env.*` patterns |
| `go.mod` / `go.sum` | Added `github.com/go-viper/mapstructure/v2` |

## Test plan

- [x] Defaults when nothing is configured -- `TestLoad_Defaults`
- [x] Real env var takes precedence over `.env` -- `TestLoad_ProcessEnvPrecedence`
- [x] `.env` value applied when variable is unset -- `TestLoad_DotEnvApplied`
- [x] Missing `.env` file is tolerated -- `TestLoad_MissingDotEnv`
- [x] Invalid config aborts startup (unparseable port, invalid env, out-of-range port) -- `TestLoad_InvalidConfig`
- [x] `.env` populates only unset keys -- `TestLoad_AdditionalScenarios`
- [x] Duration and case-insensitive log level parsing -- `TestLoad_AdditionalScenarios`


.gitignore
```diff
@@ -23,5 +23,6 @@ tmp
 @mf-types
 
 # Ignore environment variables.
-*.env*
+.env
```
YimingIsCOLD: Why change `*.env*` to `.env` and `.env.*`?

server/internal/config/config_test.go
```diff
@@ -0,0 +1,142 @@
...
+			t.Setenv(tc.key, tc.val)
+
+			_, err := config.Load()
+
+			require.HasError(t, err)
+		})
+	}
+}
+
+func TestLoad_AdditionalScenarios(t *testing.T) {
```
YimingIsCOLD: Why is this named as `AdditionalScenarios`? It seems like the test cases below have already been covered by the tests above. Double check again.

server/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+// if any value fails to parse or fails validation.
+func Load() (Config, error) {
+	cfg := defaults()
+	if err := dotenv.Load(&cfg); err != nil {
+		return Config{}, err
+	}
+	return cfg, cfg.Validate()
+}
+
+func (e Environment) Validate() error {
```
YimingIsCOLD: I think we can collapse the validation into `Config.Validate()` instead of having a special validate for the `Environment` type.

server/internal/config/config.go
```diff
@@ -0,0 +1,89 @@
...
+			WriteTimeout:      30 * time.Second,
+			IdleTimeout:       60 * time.Second,
+		},
+	}
+}
+
+// Load resolves configuration from built-in defaults, an optional .env file,
+// and real process environment variables (highest precedence). Returns an error
+// if any value fails to parse or fails validation.
+func Load() (Config, error) {
```
YimingIsCOLD: I think `Load()` should return a pointer of `Config`.

server/internal/config/config.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Load resolves configuration from built-in defaults, an optional .env file,
+// and real process environment variables (highest precedence). Returns an error
+// if any value fails to parse or fails validation.
+func Load() (*Config, error) {
+	cfg := defaults()
+	if err := dotenv.Load(&cfg); err != nil {
+		return nil, err
+	}
+	return &cfg, cfg.Validate()
```
YimingIsCOLD: It looks like if `cfg.Validate()` returns an error, we are still returning the `cfg`. When there is an error, we should return `nil` + error.

server/internal/config/config_test.go
```diff
@@ -0,0 +1,103 @@
+package config_test
```
YimingIsCOLD, line 1: Should we use a separate package for config test?

server/internal/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+// writeDotEnv writes content to a .env file in dir.
+func writeDotEnv(t *testing.T, dir, content string) {
+	t.Helper()
+	if err := os.WriteFile(filepath.Join(dir, ".env"), []byte(content), 0o644); err != nil {
+		t.Fatalf("write .env: %v", err)
+	}
+}
+
+func TestLoad_Defaults(t *testing.T) {
+	t.Run("returns built-in defaults when nothing is configured", func(t *testing.T) {
```
YimingIsCOLD: I realised that each test function contains only 1 sub-test. Do we still need to spawn sub-test?

server/internal/config/config_test.go
```diff
@@ -0,0 +1,103 @@
...
+		require.Equal(t, 60*time.Second, cfg.Server.IdleTimeout)
+	})
+}
+
+func TestLoad_ProcessEnvPrecedence(t *testing.T) {
+	t.Run("real env var takes precedence over .env file", func(t *testing.T) {
+		dir := t.TempDir()
+		writeDotEnv(t, dir, "TW_SERVER_PORT=8080")
+		chdir(t, dir)
+		t.Setenv("TW_SERVER_PORT", "9090")
```
YimingIsCOLD:
```suggestion
		t.Setenv("TW_SERVER_PORT", "9090")
		t.Cleanup(func() { _ = os.Unsetenv("TW_SERVER_PORT") }
```

We would need a cleanup function to unset whatever is in the dotenv file because dotenv package internally uses `os.Setenv` which persist across tests.

Side quest:
The exported methods in `config` package seems a bit weird. `config.Load` seems to be a wrapping the `dotenv.Load` which already tested for dotenv applied and process env precedence etc. In `teacher-workspace-archive`, the responsibility of `config` package is to return a config object with sane default values and validate method to trigger validation. It is in the main that we stitch the logic flow: `config.Default` -> `dotenv.Load` -> `config.Validate`.

server/internal/config/config_test.go
kellylimmm: I think we can have a mixture of internal and external tests.

server/internal/config/config_test.go
kellylimmm: nope, not necessary. good catch, will remove.

server/internal/config/config_test.go
kellylimmm: Setenv calls os.Setenv(key, value) and uses Cleanup to restore the environment variable to its original value after the test.
Screenshot 2026-07-06 at 2 24 21 PM

kellylimmm: Reply to side quest:

Updated to move logic so that each package has it's own responsibility and main.go stitches the whole logic together.

YimingIsCOLD (approved):

PR #48, opened 2026-07-16
feat(host/parents-gateway): render Parents Gateway remote from manifest

### LINKED ISSUE #35: feat(`host/parents-gateway`): render Parents Gateway remote from manifest
## User story

As a teacher, I want the Parents Gateway view in my workspace to render the real Parents Gateway micro-frontend, so that I can use actual Parents Gateway features instead of a static placeholder.

## Background

The host app (`apps/host`) is a Module Federation shell. It currently renders placeholder container views, including `ParentsGatewayView` ("Parents Gateway") mounted on the `/posts/*` route. The host has no remotes wired up (`remotes: {}` in `rsbuild.config.ts`) and no `loadRemote`/runtime-plugin code.

The Parents Gateway (PG) micro-frontend is now deployed, with its Module Federation manifest available at:

`https://d390008ekba73v.cloudfront.net/mf-manifest.json`

This issue wires the host to load that remote and render it in place of the `ParentsGatewayView` placeholder.

### Parents Gateway micro-frontend renders from the manifest

- **Given** the host app is running and the Parents Gateway remote is reachable at its manifest URL
- **When** a teacher navigates to the Parents Gateway route (`/posts`)
- **Then** the Parents Gateway micro-frontend loaded from the manifest is rendered in the content area, replacing the placeholder heading

### Host shell stays usable when the remote fails to load

- **Given** the Parents Gateway remote manifest cannot be loaded (for example, the manifest is unreachable)
- **When** a teacher navigates to the Parents Gateway route (`/posts`)
- **Then** a fallback state is shown in the content area and the surrounding host shell (sidebar and navigation) remains usable

## Out of scope

- The other placeholder views (`HomeView`, `StudentsView`) remain placeholders and are not wired to remotes.
- Authentication and authorization between the host and the Parents Gateway remote.

## Design assets

```mermaid
sequenceDiagram
    actor Teacher
    participant Host as Host shell
    participant CDN as CloudFront (mf-manifest.json)
    participant PG as Parents Gateway remote

    Teacher->>Host: Navigate to /posts
    Host->>CDN: Fetch mf-manifest.json
    CDN-->>Host: Manifest + remote entry
    Host->>PG: Load remote module
    PG-->>Host: Parents Gateway component
    Host-->>Teacher: Render Parents Gateway view
```

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

## Open question

- Should we conditionally load the PG view based on a given environment variable, so that during local development (when the PG view is not needed) the host continues to render the PG placeholder view?

---

---

> 🌊 Managed by **[Currents](https://tfx-currents-prod.up.railway.app)** · [Task d2ad63c7-9a86-47ad-9e32-e499e67f125a](https://tfx-currents-prod.up.railway.app/?task=d2ad63c7-9a86-47ad-9e32-e499e67f125a)

{"metadata":null,"rank":"bGHG","taskId":"d2ad63c7-9a86-47ad-9e32-e499e67f125a","version":1}

kellylimmm:
## Summary

Closes #35. Wires the host to load the Parents Gateway (PG) micro-frontend via Module Federation, replacing the static placeholder, with a fallback when the remote fails to load.

- `registerRemotes`/`loadRemote` dynamically load PG's exposed modules from its manifest at `https://d390008ekba73v.cloudfront.net/mf-manifest.json`, instead of declaring the remote statically (static declaration caused the manifest to be resolved eagerly at host bootstrap, crashing the whole app on any remote failure rather than just the `/posts` content area).
- `ErrorBoundary` + `Suspense` wrapping, scoped to just the remote content, so a failed/unreachable remote shows a fallback message while the sidebar/nav stay usable. Fallback text matches the sidebar's own labels ("Posts is unavailable right now" / "Groups is unavailable right now").
- `react-router` added as a shared singleton dependency — required for PG's `useInRouterContext()` check to correctly detect the host's router instead of mounting a second, conflicting one.
- PG's team split their single `pg/App` export into two separate federation modules (`pg/Posts`, `pg/Groups`), each with their own top-level route. Updated the host to match, with a new `/groups` route alongside `/posts`, and a Groups nav item added to the sidebar.
- All four top-level views (`HomeView`, `StudentsView`, `PostsView`, `GroupsView`) are now lazy-loaded via `React.lazy` for code-splitting. `PostsView`/`GroupsView` each independently call `registerRemotes`/`loadRemote` inline rather than sharing a helper — safe since `registerRemotes` can be called more than once for the same remote without conflict.
- Container view files named after the feature, not the underlying remote's product name, matching the existing convention: `ParentsGatewayView` → `PostsView`, `GroupsGatewayView` → `GroupsView`.

## Acceptance criteria (from #35)

- ✅ PG renders in place of the placeholder at `/posts` when reachable
- ✅ Fallback shown + host shell (sidebar/nav) stays usable when the remote fails to load

## Note

PG's own API calls use relative paths (e.g. `/api/web/2/staff/...`), so they resolve against whatever page is currently open rather than PG's real backend. This means `/posts` and `/groups` currently render PG's real components correctly, but show a "Failed to load page data" message instead of real content — expected given the above, not something this PR can fix from the host side.

## Test plan

- [x] `/posts` renders PG's real component (verified against live CloudFront deployment)
- [x] `/groups` renders PG's real component (verified against live CloudFront deployment)
- [x] Manifest/CORS failure shows fallback message, sidebar/nav remain usable
- [x] Non-200 manifest response (genuine HTTP 403) shows fallback message, sidebar/nav remain usable
- [x] `react-router` sharing confirmed necessary and working (debug-traced `useInRouterContext()` flipping from `false` to `true`)
- [x] Home/Students still render correctly after switching to lazy-loading
- [x] Typecheck and lint clean


apps/host/src/containers/GroupsGatewayView.tsx
```diff
@@ -0,0 +1,3 @@
+import { createParentsGatewayRemoteView } from '~/containers/parents-gateway-remote';
+
+export const GroupsGatewayView = createParentsGatewayRemoteView('pg/Groups', 'Groups');
```
kmye: if we are going to have app features in TW, then this should be GroupsView?

apps/host/src/containers/createParentsGatewayRemoteView.tsx
kmye, line 1: then this file will be `createParentsGatewayRemoteView.tsx` ?

apps/host/src/containers/GroupsGatewayView.tsx
kellylimmm: Good call, renamed to `GroupsView` to match the existing `HomeView/StudentsView` convention.

apps/host/src/containers/createParentsGatewayRemoteView.tsx
kellylimmm: Yep, renamed to match the exported function.

apps/host/src/containers/ParentsGatewayView.tsx
```diff
@@ -1,7 +1,3 @@
-export function ParentsGatewayView() {
-  return (
-    <div className="tw:flex tw:flex-1 tw:items-center tw:justify-center tw:p-8">
-      <h1 className="tw:text-2xl tw:font-semibold tw:text-muted-foreground">Parents Gateway</h1>
-    </div>
-  );
-}
+import { createParentsGatewayRemoteView } from '~/containers/createParentsGatewayRemoteView';
+
+export const ParentsGatewayView = createParentsGatewayRemoteView('pg/Posts', 'Parents Gateway');
```
kmye: i'm wondering if we have `GroupsViews`, then this is `PostsView` and not `ParentsGatewayView`

kellylimmm: At first i was thinking of `PostsView` too to be consistent with the rest of the naming. But i thought the [issue](https://github.com/String-dxd/teacher-workspace/issues/35) wanted it to be described as what's actually being rendered and was a bit conflicted if we should name it explicitly.

I think i'd prefer it to be `PostsView` instead for the consistency, will update to `PostsView`.

apps/host/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@base-ui/react": "^1.5.0",
     "@fontsource/inter": "^5.2.8",
+    "@module-federation/enhanced": "^2.5.1",
```
kmye, line 14: curious questions, why is this added? what are we using this for?

kellylimmm: Added this so PG only loads when someone actually hits `/posts` or `/groups`, not the second the app opens. Tried just putting it in the config directly first, but if PG ever goes down that way, it takes the whole app with it, not just those two pages. This lets us load it on demand instead, so worst case is just `Posts/Groups` break, not everything.

It's used in `createParentsGatewayRemoteView.tsx`, which both `PostsView.tsx` and `GroupsView.tsx` call into.

apps/host/src/App.tsx
```diff
@@ -3,9 +3,10 @@ import { BrowserRouter, Route, Routes } from 'react-router';
 import { AppSidebar } from '~/components/Sidebar';
 import { SidebarInset, SidebarProvider, SidebarTrigger } from '~/components/ui/sidebar';
 import { TooltipProvider } from '~/components/ui/tooltip';
+import { GroupsView } from '~/containers/GroupsView';
 import { HomeView } from '~/containers/HomeView';
 import { NotFoundView } from '~/containers/NotFoundView';
-import { ParentsGatewayView } from '~/containers/ParentsGatewayView';
+import { PostsView } from '~/containers/PostsView';
 import { StudentsView } from '~/containers/StudentsView';
```
YimingIsCOLD, line 13: We should use `React.lazy` to load the views instead of importing them directly. This will enable code-splitting for us during build.

```tsx
const HomeView = lazy(() => import('~/containers/HomeView'));
const StudentsView = lazy(() => import('~/containers/StudentsView'));
const PostsView = lazy(() => import('~/containers/PostsView'));
const GroupsView = lazy(() => import('~/containers/GroupsView'));
```

`PostsView` and `GroupsView` can include the necessary registration of modules even if both views uses the same remote module. `registerRemotes` is smart enough to only register once.

Below is an example of  `GroupsView`:
```tsx
import { loadRemote, registerRemotes } from '@module-federation/enhanced/runtime';
import React from 'react';

import { ErrorBoundary } from '~/components/ErrorBoundary';

registerRemotes([
  {
    name: 'pg',
    entry: 'https://d390008ekba73v.cloudfront.net/mf-manifest.json',
  },
]);

const RemoteApp = React.lazy(async () => {
  const module = await loadRemote<{ default: React.ComponentType }>('pg/Groups');
  if (!module) {
    throw new Error(`Failed to load pg/Groups remote module`);
  }

  return module;
});

export default function GroupsView() {
  return (
    <ErrorBoundary fallback={null}>
      <React.Suspense fallback={null}>
        <RemoteApp />
      </React.Suspense>
    </ErrorBoundary>
  );
}
```

Using the above methods, we no need to have a `createParentsGatewayRemoteView.tsx` file. We can just duplicate the fallback component in `PostsView` and `GroupsView` file.

apps/host/src/App.tsx
```diff
@@ -21,7 +22,8 @@ export default function App() {
             <Routes>
               <Route path="/" element={<HomeView />} />
               <Route path="/students/*" element={<StudentsView />} />
-              <Route path="/posts/*" element={<ParentsGatewayView />} />
+              <Route path="/posts/*" element={<PostsView />} />
+              <Route path="/groups/*" element={<GroupsView />} />
```
YimingIsCOLD: We should add the groups to the sidebar too.

kellylimmm: Added!

apps/host/src/App.tsx
kellylimmm: Sounds good, updated!

kellylimmm: Good points 👍🏻

apps/host/src/containers/PostsView.tsx
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  return module;
+});
+
+function Fallback() {
+  return (
+    <div className="tw:flex tw:flex-1 tw:items-center tw:justify-center tw:p-8">
+      <h1 className="tw:text-2xl tw:font-semibold tw:text-muted-foreground">
+        Parents Gateway is unavailable right now
```
YimingIsCOLD:
```suggestion
        Posts is unavailable right now
```

kellylimmm: Updated to `Posts`

YimingIsCOLD (approved):

apps/host/src/containers/PostsView.tsx
kellylimmm: One thing to note that although it is intended, it still sounds wrong gramatically.

Issue #44, opened 2026-07-10
feat(`server/proxy`): strip cookie and attach signed JWT on proxied requests


YimingIsCOLD:
## User story

As a downstream app backend, I want proxied API requests to arrive with a trusted, short-lived JWT identifying Teacher Workspace instead of the end user's session cookie, so that I can authenticate the caller without being exposed to the user's session credentials.

## Background

Issue #40 makes the server a transparent proxy: it forwards an MFE's API request to the matching downstream backend with method, headers, and body unchanged. That pass-through includes the `Cookie` header, which carries the Teacher Workspace session. Forwarding it leaks the user's session credential to every downstream and couples those backends to the TW session format.

Instead, just before forwarding, the proxy strips the `Cookie` header so the session credential never leaves TW, and attaches a signed JWT identifying the caller as Teacher Workspace with the target app as the audience. Claims are minimal for now: `iss` fixed to `TW`, `aud` derived from the request's app prefix, and standard `iat` and `exp`. There is no user identity in the token yet.

Signing is symmetric for now (a shared secret), with an upgrade to asymmetric signing planned later. The signing key is loaded from config and validated at server startup, so a missing key fails fast rather than at request time.

### Proxied request carries a signed JWT and no cookie

- **Given** a configured route for an app and a valid proxy signing key
- **When** an MFE sends an API request under that app's route carrying a `Cookie` header
- **Then** the forwarded request has no `Cookie` header and carries an `Authorization: Bearer <jwt>` header signed with the configured key, whose claims are `iss` = `TW`, `aud` derived from that app, and valid `iat` and `exp`

### Audience reflects the matched app

- **Given** two configured routes for two different apps, each with the signing key available
- **When** an API request is proxied under each app's route
- **Then** each forwarded request's JWT carries an `aud` claim reflecting its own app, and the two audiences differ

### Request without an incoming cookie is still signed

- **Given** a configured route and a valid signing key
- **When** an API request arrives under that route with no `Cookie` header
- **Then** the forwarded request still carries the `Authorization: Bearer <jwt>` header and reaches the downstream

## Out of scope

- Downstream verification of the JWT: each backend validates the token independently; this issue only produces and attaches it.
- Carrying user or session identity in the token (for example `sub`): claims stay minimal (`iss`, `aud`, `iat`, `exp`).
- Asymmetric signing, key rotation, and JWKS publication (future upgrade from the symmetric key).
[...]

PR #58, opened 2026-07-23
feat(`server/handler`): proxy MFE API requests to the app backends

### LINKED ISSUE #40: feat(`server/proxy`): proxy MFE API requests to downstream app backends
## User story

As an MFE app developer, I want the server to proxy my app's API requests to its downstream backend, so that browser calls stay same-origin without CORS setup or downstream URLs baked into the front end.

## Background

The host loads Module Federation remotes (for example Parents Gateway, #35), each calling its own downstream backend (Student Insight, Parents Gateway, more to come). Calling backends directly forces cross-origin requests, per-app CORS, and downstream URLs in front-end builds.

Instead, the Go server proxies API requests: an MFE calls a server API path scoped to its app, and the server forwards to the matching downstream and returns the response unchanged. Since the same server serves the host `index.html` for SPA routes (#36), an `/api` marker separates proxied API calls from client-side navigation. The app-and-`/api` prefix is stripped, so `.../api/hello` reaches the downstream as `/hello`. Routing is config-driven via the config package (#8). This issue is a transparent pass-through: no auth is injected or required.

The exact `/api` marker placement and config encoding are undecided (see Open questions); the acceptance criteria hold either way.

### Request under a configured route is proxied with its prefix stripped

- **Given** a route mapping the Student Insight app to its downstream base URL
- **When** an MFE sends an API request of any method under that app's API path, ending in `/hello`
- **Then** the app-and-`/api` prefix is stripped, the method, headers, and body are forwarded to the downstream at `/hello`, and its response status, headers, and body are returned unchanged

### Request to an unconfigured route is not proxied

- **Given** no route mapping matches the request's app segment
- **When** an API request for that unknown app arrives
- **Then** it is not forwarded and the caller receives a 404 Not Found

### Unreachable downstream surfaces as a gateway error

- **Given** a configured route whose downstream cannot be reached
- **When** an API request for that route arrives
- **Then** the proxy does not hang or crash, and the caller receives a 502 Bad Gateway

## Out of scope

- Auth and session forwarding: transparent pass-through only (see #37, #39).
- Request/response transformation, rewriting, or caching.
- Rate limiting, retries, and circuit breaking.
- WebSocket and SSE upgrade proxying (plain HTTP request/response only).
- Serving `index.html` and static assets for non-API routes (#36).

## Design assets

```mermaid
sequenceDiagram
    participant MFE as MFE app (browser)
    participant Proxy as Server proxy handler
    participant DS as Downstream backend
    MFE->>Proxy: API request under app route (.../api/hello)
    Proxy->>Proxy: Match app route in config table
    alt Route configured
        Proxy->>DS: Forward to base URL, app + /api prefix stripped (/hello)
        DS-->>Proxy: Response (status, headers, body)
        Proxy-->>MFE: Return response unchanged
    else Route not configured
        Proxy-->>MFE: 404 Not Found
    end
```

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

## Open questions

1. **API path scheme.** Global prefix `/api/student-insights/hello` (SPA fallback = anything not under `/api/`) vs marker inside the app prefix `/student-insights/api/hello` (cohesive per-app surface, but app-aware fallback). Both strip to `/hello`.
2. **Route config encoding.** A single structured env var (for example `TW_PROXY_ROUTES`), so onboarding never touches server code, vs per-downstream named vars (for example `TW_STUDENT_INSIGHTS_BASE_URL`), which need a code change per app.

---

---

> 🌊 Managed by **[Currents](https://tfx-currents-prod.up.railway.app)** · [Task 0710aecd-6bbf-4c50-a33d-84e6dbfbfef8](https://tfx-currents-prod.up.railway.app/?task=0710aecd-6bbf-4c50-a33d-84e6dbfbfef8)

{"metadata":{"sourceType":"github"},"rank":"aL","taskId":"0710aecd-6bbf-4c50-a33d-84e6dbfbfef8","version":1}

santosral:
Closes #40

## 🚀 Summary

Each MFE remote talks to its own backend. If the browser calls those backends directly, every app needs its own CORS setup, and the backend URLs get built into the frontend bundle.

This sends those calls through the Go server instead. A request to `/api/<app>/<path>` goes to that app's backend with the `/api/<app>` prefix removed. The response comes back unchanged.

## ✏️ Changes

- **`internal/config`**: `APIProxyConfig` holds one base URL per app. Both are required, so `cfg.Validate()` stops the server at startup if either is missing.
- **`internal/handler`**: `New` builds one `httputil.ReverseProxy` per app. `Register` adds the two `/api/` routes plus the `/api/` catch-all, all on the session sub-mux.
- **`internal/handler/proxy.go`** (new): the two route handlers, plus two shared handlers. `handleProxyError` writes 502 and logs the method, the path the client asked for, and the backend. The query string is left out of the log, since it can carry tokens. `handleProxyNotFound` writes the 404 for the catch-all.

## 🧪 Test Plan

`TestHandler_posts` and `TestHandler_studentInsights` cover the prefix strip, that each route reaches its own backend rather than the other one, and the 502 for a backend that is not listening. `TestHandleProxyNotFound` covers the 404. `TestHandleProxyError` checks the log records the path the client asked for rather than the stripped one, and that the query string stays out of it.


server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,107 @@
...
+	loggerFor := logger
+	if loggerFor == nil {
+		loggerFor = func(context.Context) *slog.Logger { return slog.Default() }
+	}
+
+	proxy := &httputil.ReverseProxy{
+		Transport: transport,
+		Rewrite: func(pr *httputil.ProxyRequest) {
+			pr.SetURL(target)
+			pr.SetXForwarded()
```
guangshinhaha: **[Medium] The host session cookie is forwarded verbatim to every downstream app backend**

```go
proxy := &httputil.ReverseProxy{
	Transport: transport,
	Rewrite: func(pr *httputil.ProxyRequest) {
		pr.SetURL(target)
		pr.SetXForwarded() // ← forwarding headers are handled; Cookie: tw_session=<id> rides along untouched
		if opts.ForwardedProto != "" {
			pr.Out.Header.Set("X-Forwarded-Proto", opts.ForwardedProto)
		}
	},
```

**Problem:** `ReverseProxy` only strips hop-by-hop headers, and `Cookie` is not one of them, so the host's session cookie reaches every downstream. Checked on this branch: a `GET /api/posts/hello` carrying `Cookie: tw_session=SESSION-ID-SECRET` arrives at the downstream with `Cookie: tw_session=SESSION-ID-SECRET` intact.

That session ID is the full bearer credential for the host — anything that holds one can replay it against `/` and act as that user. Per this PR, auth forwarding is out of scope (#44), and the #44 design mints a JWT from the *server-side* session, so no downstream needs the raw cookie today. Meanwhile every app backend — and its access logs and APM traces, which routinely capture request headers — accumulates live host session IDs. One backend with verbose header logging, or one operator with read access to a downstream's logs, is enough to hijack a teacher's session.

**Suggestion:** Two options.

Option A — drop cookies entirely, matching "no auth forwarding yet":

```go
Rewrite: func(pr *httputil.ProxyRequest) {
	pr.SetURL(target)
	pr.SetXForwarded()
	pr.Out.Header.Del("Cookie")
	if opts.ForwardedProto != "" {
		pr.Out.Header.Set("X-Forwarded-Proto", opts.ForwardedProto)
	}
},
```

Option B — strip only the host session cookie, so a downstream keeps any cookie it set for itself:

```go
type Options struct {
	// ...
	// SessionCookie is the host session cookie name, withheld from the
	// downstream. Empty forwards every cookie.
	SessionCookie string
}

// in Rewrite, after SetXForwarded:
if opts.SessionCookie != "" {
	inbound := pr.In.Cookies()
	pr.Out.Header.Del("Cookie")
	for _, c := range inbound {
		if c.Name != opts.SessionCookie {
			pr.Out.AddCookie(c)
		}
	}
}
```

wired from `cfg.Session.Name` in `proxyOptions`.

> A is one line and closes the hole completely, but also withholds cookies a downstream legitimately owns; B preserves those at the cost of one more option and a coupling to the session cookie name.

---
*🤖 aif-code-review · claude-opus-5[1m]*

server/internal/config/config.go
```diff
@@ -143,6 +172,58 @@ func (c SessionConfig) validate() error {
 	return errors.Join(errs...)
 }
 
+func (c APIProxyConfig) validate() error {
+	var errs []error
+
+	if c.Posts != nil {
+		if c.Posts.Scheme != "http" && c.Posts.Scheme != "https" {
+			errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must use scheme http or https; got %q", c.Posts))
```
guangshinhaha: **[Medium] A base URL with credentials is printed in the clear when validation rejects it**

```go
if c.Posts != nil {
	if c.Posts.Scheme != "http" && c.Posts.Scheme != "https" {
		errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must use scheme http or https; got %q", c.Posts)) // ←
	}
	if c.Posts.Host == "" {
		errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must include host[:port]; got %q", c.Posts)) // ←
	}
}
```

**Problem:** `%q` on a `*url.URL` routes through `url.URL.String()`, which renders userinfo in the clear. `TestAPIProxyConfig_validate/accepts valid base urls` has a `with userinfo` case, so a credentialed base URL is a supported input.

Set `TW_API_PROXY_POSTS_BASE_URL=ftp://svc:sup3rs3cret@pg.example.com` (wrong scheme — a plausible typo when someone means `http`) and `main` exits via `slog.Error("invalid config", "err", err)` with:

```
TW_API_PROXY_POSTS_BASE_URL must use scheme http or https; got "ftp://svc:sup3rs3cret@pg.example.com"
```

Ran this against `validate()` on this branch to confirm. The downstream's password lands in the startup log line and in whatever ships it onward — log aggregators are usually readable by a wider group than the config store the secret came from.

`logProxy` in `apiproxy` already reaches for `Redacted()` for exactly this reason; the config layer should match.

**Suggestion:**

```go
if c.Posts != nil {
	if c.Posts.Scheme != "http" && c.Posts.Scheme != "https" {
		errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must use scheme http or https; got %q", c.Posts.Redacted()))
	}
	if c.Posts.Host == "" {
		errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must include host[:port]; got %q", c.Posts.Redacted()))
	}
}
```

> The pre-existing `TW_DEV_SERVER_URL` checks (`config.go:105` and `108`) have the same `%q`-on-`*url.URL` shape — worth the same one-word fix while you are in here.

---
*🤖 aif-code-review · claude-opus-5[1m]*

server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,107 @@
...
+
+func logProxy(logger *slog.Logger, r *http.Request, level slog.Level, msg string, extra ...slog.Attr) {
+	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")
+	upstream := *r.URL
+	upstream.RawQuery = ""
+
+	logger.LogAttrs(r.Context(), level, msg, append([]slog.Attr{
+		slog.String("method", r.Method),
+		slog.String("path", inboundPath),
+		slog.String("upstream", upstream.Redacted()),
```
guangshinhaha: **[Low] The `upstream` log attribute is a relative path on two of the three call sites**

```go
func logProxy(logger *slog.Logger, r *http.Request, level slog.Level, msg string, extra ...slog.Attr) {
	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")
	upstream := *r.URL
	upstream.RawQuery = ""

	logger.LogAttrs(r.Context(), level, msg, append([]slog.Attr{
		slog.String("method", r.Method),
		slog.String("path", inboundPath),
		slog.String("upstream", upstream.Redacted()), // ← only absolute when r is the outbound request
	}, extra...)...)
}
```

**Problem:** `r.URL` is an absolute downstream URL only when `logProxy` is called from `ErrorHandler` — `ReverseProxy` hands that callback `outreq`, whose URL the `Rewrite` hook already rewrote. The two call sites inside the returned handler (`apiproxy.go:59` "refused a streaming request" and `apiproxy.go:71` "truncated the response mid-body") pass the *inbound* request, whose `URL` is a bare path that `http.StripPrefix` has already stripped of `/api/<app>`.

Ran it on this branch: an SSE request to `/api/posts/hello?x=1` logs

```json
{"level":"WARN","msg":"refused a streaming request","method":"GET","path":"/hello","upstream":"/hello"}
```

So `upstream` duplicates `path` instead of naming a backend. It matters most for `"truncated the response mid-body"` — the one line that tells an operator a downstream cut a response short — where the field that should identify the offending backend identifies nothing. With more than one app configured, that line cannot be attributed.

**Suggestion:** Pass the upstream explicitly rather than deriving it from whichever request happens to be in hand:

```go
func logProxy(logger *slog.Logger, r *http.Request, upstream *url.URL, level slog.Level, msg string, extra ...slog.Attr) {
	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")
	u := *upstream
	u.RawQuery = ""

	logger.LogAttrs(r.Context(), level, msg, append([]slog.Attr{
		slog.String("method", r.Method),
		slog.String("path", inboundPath),
		slog.String("upstream", u.Redacted()),
	}, extra...)...)
}
```

Call sites:

```go
// ErrorHandler — r is the outbound request, so its URL is already the upstream.
logProxy(loggerFor(r.Context()), r, r.URL, slog.LevelError, "failed to proxy request", slog.Any("err", err))

// Handler — r is inbound, so resolve against the target this proxy was built for.
logProxy(loggerFor(r.Context()), r, target.JoinPath(r.URL.EscapedPath()), slog.LevelWarn, "refused a streaming request")
logProxy(loggerFor(ctx), r, target.JoinPath(r.URL.EscapedPath()), slog.LevelError, "truncated the response mid-body")
```

---
*🤖 aif-code-review · claude-opus-5[1m]*

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 2 |
| Low      | 1 |

## What Looks Good

- **The `/api/` 404 floor next to the per-app prefixes closes the fall-through properly.** An unconfigured app can never land on the SPA index or the dev-server proxy, in either environment — and `http.StripPrefix`'s `RawPath` check makes `/api/post%73/hello` fail closed instead of normalising into a live route. Both directions are pinned by tests.
- **Both proxy timeouts are validated strictly below `TW_SERVER_WRITE_TIMEOUT` at startup**, and `TestNew_WriteTimeoutInteraction` reads the literal `HTTP/1.1 502 Bad Gateway` status line off a real socket rather than asserting on a recorder. That is the right level to test a write-deadline interaction, and it catches the class of bug where the deadline eats the error response.
- **`clientGone` keys off the request context rather than error-string matching**, so a caller hanging up mid-upload stays out of the error log while a refused dial and a header-deadline miss both stay in it. `TestClientGone` pins all six cases, including the `*url.Error` wrapper the transport adds.
- **Query strings are stripped from proxy error logs**, with `omits the query string from the gateway error log` asserting that neither a token nor a student ID reaches the log — the right default for this product, decided where the log is written rather than left to a downstream scrubber.

---
*🤖 aif-code-review · claude-opus-5[1m]*

server/internal/config/config.go
```diff
@@ -143,6 +172,58 @@ func (c SessionConfig) validate() error {
...
+		}
+		if c.Posts.Host == "" {
+			errs = append(errs, fmt.Errorf("TW_API_PROXY_POSTS_BASE_URL must include host[:port]; got %q", c.Posts))
+		}
```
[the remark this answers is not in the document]
santosral: Fixed, but not for the reason given. `*url.URL` has `String()` on a pointer
receiver, so `%q` uses it and prints a normal quoted URL. There is no `%!q`
placeholder here.

The real problem is that the whole URL prints, password included:

    TW_API_PROXY_POSTS_BASE_URL must use scheme http or https; got "ftp://svc:sup3rs3cret@pg.example.com"

Both checks now use `.Redacted()`, with a test that fails if the password comes
back. Folded into 4003262.

server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,107 @@
...
+				if recovered == http.ErrAbortHandler {
+					logProxy(loggerFor(ctx), r, slog.LevelError, "truncated the response mid-body")
+				}
+
+				panic(recovered)
+			}
+		}()
+
+		proxy.ServeHTTP(w, r.WithContext(ctx))
+	})
```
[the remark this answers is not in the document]
santosral: This line was already correct. `ReverseProxy` hands the outbound request to
`ErrorHandler`, and `Rewrite` has set Scheme and Host on it by then, so the log
does name the backend:

    "msg":"failed to proxy request","upstream":"http://127.0.0.1:1/v1/hello"

The gap was on the other `logProxy` calls, which pass the inbound request and
logged a bare path. `logProxy` now takes the upstream as an argument, with two
new tests. Folded into 610e8c6 and e83a239.

server/internal/apiproxy/apiproxy.go
santosral: Agreed. This is fixed in #69, which is stacked on this branch. It drops the
Cookie header and sends a signed JWT instead, so no downstream sees the session
ID.

Checked end to end: the downstream now gets an empty Cookie header and an
Authorization: Bearer header.

One note on ordering. If this PR merges first, main forwards session cookies
until #69 lands.

server/internal/config/config.go
santosral: Fixed with `.Redacted()` on both checks, plus a test that fails if the password
comes back. Folded into 4003262.

Leaving the TW_DEV_SERVER_URL pair for a separate PR. Those lines are older than
this branch and sit outside #40, so they do not belong in these commits.

server/internal/apiproxy/apiproxy.go
santosral: Fixed, taking the suggested approach. `logProxy` now takes the upstream as an
argument instead of reading it off whichever request is in hand.

Two notes from checking it.

`StripPrefix` does not touch `RequestURI`, so `path` kept the full inbound path
and `upstream` was the stripped one:

    "msg":"refused a streaming request","path":"/api/posts/hello","upstream":"/hello"

So they were not duplicates, but the point stands: upstream named no backend.

The truncation path also had no test at all. `ReverseProxy` only raises
`ErrAbortHandler` when a real `http.Server` is in the request context, and every
test here drove the handler directly, so deleting the whole recover block kept
the suite green. Both call sites are now pinned by tests that run the proxy
behind a real server.

Folded into 610e8c6 and e83a239.

server/internal/config/config.go
```diff
@@ -68,6 +82,12 @@ func Default() Config {
 			DefaultTTL:       3 * time.Hour,
 			AuthenticatedTTL: 30 * time.Minute,
 		},
+		APIProxy: APIProxyConfig{
+			ResponseHeaderTimeout: 10 * time.Second,
```
YimingIsCOLD: I think the default can be the same as ServerConfig's `ResponseHeaderTimeout`?

server/internal/config/config.go
```diff
@@ -47,6 +50,17 @@ type SessionConfig struct {
...
 
+// APIProxyConfig represents the configuration for the downstream API proxy.
+type APIProxyConfig struct {
+	ResponseHeaderTimeout time.Duration `dotenv:"TW_API_PROXY_RESPONSE_HEADER_TIMEOUT"`
+	RequestTimeout        time.Duration `dotenv:"TW_API_PROXY_REQUEST_TIMEOUT"`
+	MaxConnsPerHost       int           `dotenv:"TW_API_PROXY_MAX_CONNS_PER_HOST"`
+	MaxIdleConnsPerHost   int           `dotenv:"TW_API_PROXY_MAX_IDLE_CONNS_PER_HOST"`
+	ForwardedProto        string        `dotenv:"TW_API_PROXY_FORWARDED_PROTO"`
+
+	Posts *url.URL `dotenv:"TW_API_PROXY_POSTS_BASE_URL"`
```
```suggestion
	PostsBaseURL *url.URL `dotenv:"TW_API_PROXY_POSTS_BASE_URL"`
```

server/internal/apiproxy/apiproxy.go
YimingIsCOLD: I think we can create an issue for all the forwarded proto stuff as we need more information to make decisions on whether to forward the TW backend IP or client IP to the downstream backend.

server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,107 @@
...
+			if !clientGone(r, err) {
+				logProxy(loggerFor(r.Context()), r, r.URL, slog.LevelError, "failed to proxy request", slog.Any("err", err))
+			}
+
+			http.Error(w, "Bad Gateway", http.StatusBadGateway)
+		},
+	}
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if streaming(r) {
```
YimingIsCOLD: I think we don't need to block streaming for now. By right the proxy should be able to proxy streams already. We can have another issue to track if stream works properly.

server/internal/config/config.go
```diff
@@ -143,6 +172,58 @@ func (c SessionConfig) validate() error {
...
+	}
+	if c.ForwardedProto != "" && c.ForwardedProto != "http" && c.ForwardedProto != "https" {
+		errs = append(errs, fmt.Errorf("TW_API_PROXY_FORWARDED_PROTO must be http or https when set; got %q", c.ForwardedProto))
+	}
+
+	return errors.Join(errs...)
+}
+
+// Targets maps each configured app to its downstream base URL, omitting unset apps.
+func (c APIProxyConfig) Targets() map[string]*url.URL {
```
YimingIsCOLD: This kind of business logic can be in a `proxy.go` in `handler` package.

```go
func (h *Handler) proxy(w http.ResponseWriter, r *http.Request) {
      app := r.PathValue("app")

      var p *httputil.ReverseProxy
      switch app {
      case "posts":
              p = h.postsProxy
      }
      if p == nil {
              logger := middleware.LoggerFromContext(r.Context())
              httputil.RenderPlain(w, logger, http.StatusNotFound)
              return
      }

      http.StripPrefix("/api/"+app, p).ServeHTTP(w, r)
}
```

server/internal/config/config.go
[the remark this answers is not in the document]
YimingIsCOLD: We don't need `.Redacted()` because there won't be a scenario that we will do HTTP basic authentication with downstream backend.

server/cmd/tw/main.go
```diff
@@ -59,7 +59,7 @@ func main() {
 	defer stop()
 
 	go func() {
-		slog.Info("listening", "addr", addr, "env", cfg.Env)
+		slog.Info("listening", "addr", addr, "env", cfg.Env, "api_proxy_apps", cfg.APIProxy.Apps())
```
YimingIsCOLD: We don't need to log the proxies.

server/internal/config/config.go
santosral: Removed this validation. Not necessary.

server/internal/apiproxy/apiproxy.go
santosral: Removed, refactored the code.

server/internal/handler/handler.go
```diff
@@ -41,5 +58,10 @@ func (h *Handler) Register(mux *http.ServeMux, session middleware.Middleware) {
 	app := http.NewServeMux()
 	app.HandleFunc("/", h.index)
 
+	app.HandleFunc("/api/{app}/", h.proxy)
+	app.HandleFunc("/api/", func(w http.ResponseWriter, r *http.Request) {
+		httputil.RenderPlain(w, middleware.LoggerFromContext(r.Context()), http.StatusNotFound)
```
santosral: Use RenderJSON().

YimingIsCOLD: Done that!

YimingIsCOLD (approved):

PR #59, opened 2026-07-27
feat(`server/session`): add stateful session middleware

### LINKED ISSUE #37: feat(`server/session`): add stateful session middleware with in-memory store
## Background

Server-side sessions are a prerequisite for Edupass sign-in (#12). We need the middleware and a `Store` abstraction now, with an in-memory store sufficient for local development and tests. The Valkey-backed store is deferred. A reference implementation exists in the `teacher-workspace-archive` repo (`server/internal/session/session.go` and `server/internal/session/memstore/memstore.go`).

### Session is created for a request with no session cookie

- **Given** an HTTP request with no session cookie
- **When** it passes through the session middleware
- **Then** a new empty session is created, exposed on the request context, and a session cookie is set on the response

### Session is loaded for a request with a known session cookie

- **Given** an HTTP request whose session cookie references a session held in the store
- **When** it passes through the session middleware
- **Then** the stored session is loaded and exposed to downstream handlers via the request context

### Session changes are persisted to the store

- **Given** a handler that writes a value into the request's session
- **When** the request completes
- **Then** the value is saved and is readable on a later request carrying the same cookie

### Unknown or expired session cookie yields a fresh session

- **Given** an HTTP request whose session cookie references an ID not present in the store
- **When** it passes through the session middleware
- **Then** a new empty session is created (no error) and a new session cookie is issued

## Out of scope

- Valkey/Redis-backed session store (separate issue)
- CSRF enforcement middleware (the session carries a CSRF token, but validating it is separate)
- Cross-process/shared session state (in-memory store is per-process only)

## Design assets

N/A

---

## Technical context

New `server/internal/session` package owns the `Session`/`Snapshot` types and `Store` interface; the in-memory store lives under `server/internal/session/memstore`. Mirror the archive reference for types, store interface, and TTL semantics.

The middleware wrapper lives in `server/internal/middleware` (e.g. `session.go`), following `requestid.go`: wrap `http.Handler`, stash the session on the request context behind an unexported `ctxKey`, expose `SessionFromContext(ctx)`. Per-request flow: read the cookie, `Store.Prepare(ctx, id)` (on `(nil, nil)` use `session.New()`, else `session.FromSnapshot`), serve, then `Store.Commit(ctx, sess.Snapshot(), ttl)`; if the ID rotated mid-request, `Drop` the old entry and re-cookie.

Cookie `Secure` and the session TTL come from the config package (#8), not `os.Getenv`.

Patterns: `server/internal/middleware/requestid.go`, `requestid_test.go`, `server/pkg/random/random.go`.

## Data model

Mirror the archive's shapes (same names/semantics, do not copy verbatim):

- `session.Session`: per-request state (id, CSRF token, optional user, `data map[string]any`), single-goroutine owned. Accessors `ID`/`CSRFToken`/`User`/`IsAuthenticated`, data helpers `Get`/`Set`/`Delete`, `SetUser` (rotates ID + CSRF token on unauthenticated->authenticated to prevent fixation), `Rotate`.
- `session.Snapshot`: storage form (`ID`, `CSRFToken`, optional `User`, `Data`) with JSON tags; converted via `New`/`FromSnapshot`/`Snapshot()` (shares the `Data` map).
- `session.User`: `Email string`.
- `session.Store`: `Prepare(ctx, id) (*Snapshot, error)` ((nil, nil) when absent/expired), `Commit(ctx, *Snapshot, ttl time.Duration) error`, `Drop(ctx, id) error`.
- `memstore.Store`: `map[string]entry` (snapshot + `expiresAt`) under a `sync.Mutex`, lazy TTL eviction on read, `WithClock` for tests.

## API contract

No new HTTP route. Session cookie (`Set-Cookie`): name `tw_session`, value `pkg/random.Base58(32)`, `HttpOnly`, `SameSite=Lax`, `Path=/`, `Secure` off when `TW_ENV == "development"` else on, `Max-Age` = configured TTL.

Go surface: `session.{New, FromSnapshot, (*Session).Snapshot, Store}` (+ methods above); `memstore.{New(opts...), WithClock}`; a `middleware` session constructor taking a `session.Store` plus cookie/TTL config, and `SessionFromContext(ctx) (*session.Session, bool)`.

## Error contract

- Missing/unknown/expired cookie: `Prepare` returns `(nil, nil)`; create a fresh session and cookie. Not an HTTP error.
- `Commit` arg violations (nil snapshot, empty ID, non-positive TTL) are programmer errors; the middleware always passes valid args.
- Unexpected store error: log via `middleware.LoggerFromContext(ctx)` and continue with a new empty session, no 5xx.

### Concurrent store access is race-free

- **Given** the in-memory store
- **When** goroutines call `Prepare`/`Commit`/`Drop` concurrently
- **Then** no data races under `go test -race` and no panics

### Expired entries are evicted on read

- **Given** a committed session past its TTL (clock advanced via `WithClock`)
- **When** `Prepare` is called for its ID
- **Then** it returns `(nil, nil)` and the entry is removed

### ID rotation drops the old entry and re-cookies

- **Given** a request whose handler attaches a user to a previously unauthenticated session
- **When** the request completes
- **Then** the old ID is dropped, the new ID committed, and the response cookie carries the new ID

## Hard constraints

- No new third-party Go dependency: stdlib plus existing `pkg/random` and config packages only.
- No Valkey/Redis store in this issue.
- The in-memory store must be concurrency-safe; `Session` itself stays single-goroutine owned.
- Read `TW_ENV` and the TTL from the config package (#8), not env vars directly.
- Match the reference package's exported names/semantics; do not change `requestid`/`requestlog` behavior.

---

*🤖 Groomed with aif-groom-issue*

---

> 🌊 Managed by **[Currents](https://tfx-currents-prod.up.railway.app)** · [Task e36fca89-07a3-43a5-b8e8-45997e7bf300](https://tfx-currents-prod.up.railway.app/?task=e36fca89-07a3-43a5-b8e8-45997e7bf300)

{"metadata":{"sourceType":"github"},"rank":"aL","taskId":"e36fca89-07a3-43a5-b8e8-45997e7bf300","version":1}

YimingIsCOLD:
Closes #37 

## 🚀 Summary

Gives the server stateful sessions. Every request carries a session cookie backed by a pluggable store, and handlers read and mutate a typed `*session.Session` from the request context without ever touching the cookie themselves.

The part worth reviewing closely is **when** the cookie is written. `http.SetCookie` only writes into the header map, and that map goes to the wire on the handler's first write. Setting the cookie before the handler runs therefore freezes the pre-handler session ID into the response, so a future `POST /login` handler calling `SetUser` would rotate the ID (the session-fixation defence) but ship the browser the old one. The middleware instead defers the cookie **and** the store write to the handler's first write, through a `ResponseWriter` wrapper. A login handler is then just:

```go
sess, _ := middleware.SessionFromContext(r.Context())
sess.SetUser(&session.User{Email: email})
http.Redirect(w, r, "/", http.StatusSeeOther)
```

The contract this creates: **mutate the session before writing the response.** Anything set after the first byte reaches neither the cookie nor the store. This is documented on `Session`.

## ✏️ Changes

- **`internal/middleware`**: `Session` middleware that loads the session from the request cookie, refreshes the cookie with a sliding `Max-Age`, and commits to the store once the handler is done. Rotation drops the entry it superseded, committing before dropping so a failed commit can't leave a client holding a cookie for an entry that no longer exists. Adds the `Middleware` type.
- **`internal/session/memstore`**: entries now hold marshalled JSON instead of a `*session.Snapshot`. `Prepare` previously returned the stored snapshot and `FromSnapshot` shares its `Data` map, so two concurrent requests on one cookie held a single map between them and a `Set` from each was a concurrent map write. Mutations after `Commit` also leaked back into the store, which a serializing store would have dropped.
- **`internal/config`**: `TW_SESSION_NAME`, `TW_SESSION_DEFAULT_TTL`, `TW_SESSION_AUTHENTICATED_TTL` with defaults and validation. Also flattens `TW_HOST_DEV_SERVER_URL` / `TW_HOST_BUILD_DIR` to `TW_DEV_SERVER_URL` / `TW_BUILD_DIR`, since the server only ever serves the one frontend.
- **`internal/handler`**: `New` returns `*Handler` and `Register` wires routes onto a mux, splitting the former `ServeHost` into `index` and `static`. Static assets are registered outside the session middleware, so asset responses carry no session cookie and skip the session store.
- **`internal/httputil`**: shared header/MIME constants and `RenderPlain`.
- **Naming**: `responseRecorder` becomes `requestLogResponseWriter`, matching `sessionResponseWriter`, so both wrappers read as `<middleware>ResponseWriter`.

Split into 8 commits, each of which builds and tests green on its own.

## 🧪 Test Plan

Automated only. There's no login handler yet, so rotation is exercised through tests rather than a browser.

- `go test -race ./server/...`, `make lint` (0 issues) and `gofmt -l server/` all clean.
- Every commit checked out in a scratch worktree and verified to build and pass `go test ./server/...` in isolation.

New middleware tests: rotation writes the new ID with the authenticated `Max-Age` and drops the old entry; no drop without rotation; no drop when the commit fails; drop failures logged without touching the response; one save across multiple writes; post-write mutations reach neither cookie nor store.

New memstore tests: independent snapshot per caller; committed state unreachable from the caller afterwards; a `chan` in `Data` rejected with nothing stored; undecodable entry surfaces an error; and a round-trip test pinning that an `int` returns as a `float64`, so the JSON coercion is documented behaviour rather than a surprise.

Both central behaviours were verified to fail without their fix, not just asserted:

- Flipping the save to run *after* `WriteHeader` makes the rotation test fail with `want session cookie to be set`, because `httptest` snapshots headers at `WriteHeader` exactly as `net/http` flushes them.
- The memstore aliasing race was reproduced under `-race` before the change, reported at `session.(*Session).Set` with two goroutines writing the same map.


server/internal/config/config.go
```diff
@@ -16,20 +16,21 @@ const (
...
-
-	// DevServerURL is where the Go server proxies to in development.
-	DevServerURL *url.URL `dotenv:"TW_HOST_DEV_SERVER_URL"`
-	// BuildDir is where the Go server reads apps/host's build output from in production.
-	BuildDir string `dotenv:"TW_HOST_BUILD_DIR"`
+	Env      Environment `dotenv:"TW_ENV"`
+	LogLevel slog.Level  `dotenv:"TW_LOG_LEVEL"`
+
+	// DevServerURL is used in development to proxy requests to the frontend development server.
+	DevServerURL *url.URL `dotenv:"TW_DEV_SERVER_URL"`
```
santosral, line 26: Dropping the `HOST_` infix reads well. This server only ever serves the one frontend, so the extra namespace wasn't carrying meaning, and `.env.example` came along with the rename.

The four `Validate()` messages didn't:

https://github.com/String-dxd/teacher-workspace/blob/e6ca949958e753492f74080e5f4d0bcb8c0f2c3d/server/internal/config/config.go#L83-L94

```console
$ TW_DEV_SERVER_URL=ftp://x go run ./server/cmd/tw
ERROR invalid config err="TW_HOST_DEV_SERVER_URL must use scheme http or https; got \"ftp://x\""
```

Worth knowing why nothing caught it: `TestValidate` already exercises all four of those branches, but every case ends at `require.HasError(t, cfg.Validate())`, which checks that an error happened and never what it said. So a tag and its message can drift apart with the suite fully green. Asserting on the substring in one case would pin them together, if you think it's worth the line.

santosral (changes requested):

server/internal/middleware/session.go
```diff
@@ -0,0 +1,170 @@
...
+				// The save can outlive the request context, which a client that
+				// disconnects mid-response cancels.
+				ctx := context.WithoutCancel(r.Context())
+
+				// The response is already in flight, so the status code cannot
+				// be changed here. Log failures and return.
+				//
+				// Commit before Drop: a failed commit must not leave the client
+				// holding a cookie for an entry that no longer exists.
+				if err := store.Commit(ctx, sess.Snapshot(), ttl); err != nil {
```
santosral: The comment directly above this line states the invariant, and the ordering does not hold it.

`http.SetCookie` already ran at [`session.go:118`](https://github.com/String-dxd/teacher-workspace/blob/e6ca949958e753492f74080e5f4d0bcb8c0f2c3d/server/internal/middleware/session.go#L118), nineteen lines up. When `Commit` fails the code logs and returns, leaving that cookie in the header map, so the client walks away holding an ID that was never stored. On a rotating request it also returns before the `Drop`, so the superseded entry outlives the one meant to replace it.

It fires today. `Set(key string, val any)` accepts anything and `memstore.Commit` marshals, so a single non-marshalable value does it:

```console
ERROR failed to commit session err="memstore: marshal snapshot: json: unsupported type: func()"
status=200 cookies=1
client got qbHefgcL...; store lookup snap=<nil> err=<nil>
```

A teacher posts to a future `/login`, gets a 200 and a cookie, and is anonymous on the very next request, with no way back to the old session either. Both calls happen before anything reaches the wire, so the order is free: commit first, set the cookie only on success.

`session_test.go`'s "preserves the response and logs when Commit fails" checks the status, the body and the log line, but never the cookie, which is why this passes today.

server/internal/middleware/session.go
```diff
@@ -0,0 +1,170 @@
...
+// saveOnce runs the save hook at most once per request.
+func (rw *sessionResponseWriter) saveOnce() {
+	rw.once.Do(func() { rw.save(rw.ResponseWriter) })
+}
+
+// WriteHeader saves the session, then forwards to the underlying ResponseWriter.
+func (rw *sessionResponseWriter) WriteHeader(status int) {
+	rw.saveOnce()
+	rw.ResponseWriter.WriteHeader(status)
+}
```
santosral, line 39: `WriteHeader` treats every status as "the headers are about to be committed" and burns the `sync.Once`. A 1xx does not commit them, and `httputil.ReverseProxy` clears the header map immediately after forwarding one ([`reverseproxy.go:556-557`](https://github.com/golang/go/blob/go1.26.1/src/net/http/httputil/reverseproxy.go#L556-L557)). So the hook commits the session and writes `Set-Cookie` into a map that is then wiped, and the `Once` guarantees there is no second attempt: the client gets no cookie and the store keeps an entry nobody can reach.

Skipping the save for `100 <= status < 200` closes it.

Correcting the "no impact today" note I first left here: it does fire on the dev stack as it stands. The rsbuild dev server is a Node `http` server, and Node auto-answers `100 Continue` when a request carries `Expect: 100-continue` and nothing registered a `checkContinue` listener. A request body of unknown length gets that header, so `curl -T -` is enough:

```console
$ pnpm dev &
$ TW_ENV=development TW_SERVER_PORT=4321 TW_DEV_SERVER_URL=http://127.0.0.1:3001 go run ./server/cmd/tw &
$ head -c 5000 /dev/zero | tr '\0' 'x' | curl -sv -X POST -T - http://127.0.0.1:4321/ -o /dev/null
< HTTP/1.1 100 Continue
< Set-Cookie: tw_session=uuCQCVjUdG9Vrq3FE1JyjXyezSkDJozA; Path=/; Max-Age=10800; HttpOnly; SameSite=Lax
< X-Request-Id: y5kpQWHrzkFqBUerTZB8wFPfnFAUwPrD
<
< HTTP/1.1 404 Not Found
< Content-Type: text/plain; charset=utf-8
< Vary: Origin
< Transfer-Encoding: chunked
```

The cookie rides the `100`, the `404` carries none, and `X-Request-Id` from `RequestLog` goes with it. With the guard applied both land on the `404` instead, and `go test -race ./server/...` stays green.

Still narrow enough that I would not call it a blocker. Browsers do not send `Expect`, and a sized body does not either at any size (`--data-binary @file` at 100kB sends no `Expect` on curl 8.7.1), so ordinary dev traffic is unaffected. It gets more reachable once #58's `/api/` proxy sits under this middleware and real clients start POSTing to downstream apps.

Worth recording that the guard cannot regress the upgrade path, which was my first worry about adding it. Go's transport treats 101 as terminal ([#26161](https://github.com/golang/go/issues/26161)) so it never reaches `Got1xxResponse`, and `ReverseProxy` hijacks the connection for upgrades rather than calling `WriteHeader`. Proxying a websocket handshake through a wrapper carrying the guard records zero `WriteHeader` calls, completes the 101, and still round-trips bytes afterwards. HMR never touches this branch.

server/internal/middleware/session.go
```diff
@@ -0,0 +1,170 @@
...
+			rw := &sessionResponseWriter{ResponseWriter: w}
+			rw.save = func(w http.ResponseWriter) {
+				ttl := opts.DefaultTTL
+				if sess.IsAuthenticated() {
+					ttl = opts.AuthenticatedTTL
+				}
+
+				// Refresh the cookie on every request so its Max-Age slides
+				// with the store TTL.
+				http.SetCookie(w, &http.Cookie{
```
santosral, line 137: This is the one I would fix before merge, even though nothing exploits it today.

Session-bearing responses carry no cache directive. A 200 is heuristically cacheable on its own ([RFC 9110 section 15.3.1](https://www.rfc-editor.org/rfc/rfc9110#section-15.3.1)), so a shared cache may store this response with the cookie attached ([RFC 9111 section 7.3](https://www.rfc-editor.org/rfc/rfc9111#section-7.3): "a cacheable response with a Set-Cookie header field can be (and often is) used to satisfy subsequent requests to caches").

```console
$ curl -sD- -o /dev/null http://127.0.0.1:4123/ | grep -iE '^HTTP|^Cache-Control|^Vary|^Set-Cookie'
HTTP/1.1 200 OK
Set-Cookie: tw_session=AbEAFt5B...; Path=/; Max-Age=10800; HttpOnly; Secure; SameSite=Lax
```

The grep asks for `Cache-Control` and `Vary`; neither comes back.

Put a CDN or a TLS-terminating proxy in front and the first visitor's `Set-Cookie` can be served to everyone behind them, so they share one session ID. Two cookie-less visitors share a cache key too, since an absent header only matches another absent one ([RFC 9111 section 4.1](https://www.rfc-editor.org/rfc/rfc9111#section-4.1)). No shared cache in the repo today and no login handler yet, so this is about closing the seam while it is two lines:

```go
w.Header().Add("Vary", "Cookie")
w.Header().Add("Cache-Control", "no-store")
```

`no-store` is doing the work here and `Vary` is defence in depth. Worth knowing scs takes the narrower `no-cache="Set-Cookie"` ([session.go#L217](https://github.com/alexedwards/scs/blob/ab20b3feb5e9981c1f79cee8a97a289810134163/session.go#L217)), which would keep the shell cacheable, but the RFC notes the qualified form "is not widely implemented" and is often handled as a bare `no-cache` ([section 5.2.2.4](https://www.rfc-editor.org/rfc/rfc9111#section-5.2.2.4)), which still stores the cookie. I would take `no-store`.

#37 enumerated the cookie attributes precisely and said nothing about cache headers, so this is a gap in the issue as much as in the code.

server/internal/config/config.go
```diff
@@ -109,3 +118,27 @@ func (c ServerConfig) validate() error {
 
 	return errors.Join(errs...)
 }
+
+func (c SessionConfig) validate() error {
+	var errs []error
+
+	if c.Name == "" {
+		errs = append(errs, errors.New("TW_SESSION_NAME is required"))
+	}
```
santosral: Worth a charset check while you are in here.

`validate()` only rejects the empty string, but `http.SetCookie` drops the header outright when the name is not a valid token, and says so itself: "The provided cookie must have a valid Name. Invalid cookies may be silently dropped" ([`cookie.go:248-255`](https://github.com/golang/go/blob/go1.26.1/src/net/http/cookie.go#L248-L255)).

```console
name="tw_session"   status=200 setCookie=1
name="tw session"   status=200 setCookie=0
name="tw=session"   status=200 setCookie=0
name="tw;session"   status=200 setCookie=0
```

An operator who puts `TW_SESSION_NAME=tw session` in `.env` gets a clean startup, 200s everywhere, no session for anyone, and a 3h store entry per request.

server/internal/handler/handler.go
```diff
@@ -29,24 +27,20 @@ func New(cfg *config.Config) http.Handler {
...
-		if strings.HasPrefix(r.URL.Path, "/static/") {
-			h.assets.ServeHTTP(w, r)
-			return
-		}
-		http.ServeFile(w, r, filepath.Join(h.cfg.BuildDir, "index.html"))
-	}
+// Register registers all application routes on the given HTTP server mux. The
+// session middleware is applied to the application routes but not to static
+// assets, so asset responses carry no session cookie and skip the session store.
+func (h *Handler) Register(mux *http.ServeMux, session middleware.Middleware) {
```
santosral, line 36: Worth closing before this merges: the session middleware can be unwired from every application route and the suite stays green.

```console
$ # in Register: mux.Handle("/", session(app))  ->  mux.Handle("/", app)
$ go test ./... -count=1
(every package ok)
```

`New` returning `*Handler` rather than `http.Handler` is what took `handler_test.go` out, since `mux.ServeHTTP` no longer compiled, and calling `h.index` and `h.static` directly is a reasonable thing to reach for instead. It does move the suite from routing to handlers, though. `GET /static/js/missing.js` returning 404, `GET /dashboard` falling back to index.html and `POST /` doing the same went with it, and `Register` is now the only function here that no test calls. The new `default:` branch cases are a real gain in the other direction.

Two lines put the mux back under test:

```go
mux := http.NewServeMux()
handler.New(&cfg).Register(mux, sessionMW)
```

Status and body will not separate the two routes, which is the part worth knowing before writing it: in development `index` and `static` are the same statement ([`index.go:18`](https://github.com/String-dxd/teacher-workspace/blob/e6ca949958e753492f74080e5f4d0bcb8c0f2c3d/server/internal/handler/index.go#L18), [`index.go:33`](https://github.com/String-dxd/teacher-workspace/blob/e6ca949958e753492f74080e5f4d0bcb8c0f2c3d/server/internal/handler/index.go#L33)), both proxying to rsbuild. A `Prepare` counter tells them apart: one `/static/...` asserting no cookie and zero prepares, one `/dashboard` asserting both.

server/internal/config/config.go
```diff
@@ -109,3 +118,27 @@ func (c ServerConfig) validate() error {
...
+	if c.DefaultTTL <= 0 {
+		errs = append(errs, fmt.Errorf("TW_SESSION_DEFAULT_TTL must be positive duration; got %v", c.DefaultTTL))
+	}
+	if c.AuthenticatedTTL <= 0 {
+		errs = append(errs, fmt.Errorf("TW_SESSION_AUTHENTICATED_TTL must be positive duration; got %v", c.AuthenticatedTTL))
+	}
```
santosral: Nit. `Max-Age` is whole seconds, so `int(ttl.Seconds())` truncates. Below a second that truncates to zero, and `net/http` reads `MaxAge == 0` as "omit the attribute" rather than "expire now" ([`cookie.go:299-304`](https://github.com/golang/go/blob/go1.26.1/src/net/http/cookie.go#L299-L304)). So a sub-second TTL passes `Validate()` and ships a cookie with neither `Max-Age` nor `Expires`, which is a browser-session cookie: it outlives its 500ms store entry until the browser closes.

```console
$ pnpm build   # production mode validates that TW_BUILD_DIR exists
$ TW_ENV=production TW_SESSION_DEFAULT_TTL=500ms TW_SERVER_PORT=4131 go run ./server/cmd/tw &
$ curl -sD- -o /dev/null http://127.0.0.1:4131/ | grep Set-Cookie
Set-Cookie: tw_session=iQh9r48r...; Path=/; HttpOnly; Secure; SameSite=Lax
```

Nobody deploys a 500ms session, so this only earns a line because #37 pins "`Max-Age` = configured TTL".

```suggestion
	if c.DefaultTTL < time.Second {
		errs = append(errs, fmt.Errorf("TW_SESSION_DEFAULT_TTL must be at least 1s; got %v", c.DefaultTTL))
	}
	if c.AuthenticatedTTL < time.Second {
		errs = append(errs, fmt.Errorf("TW_SESSION_AUTHENTICATED_TTL must be at least 1s; got %v", c.AuthenticatedTTL))
	}
```

Build, tests and lint stay green with that applied. It closes the vanishing-attribute case only: `1500ms` still passes and still ships `Max-Age=1`, so cookie and store stay half a second apart. Rounding is a separate call and I would leave it.

server/internal/config/config.go
YimingIsCOLD: Fixed in `cbae3c3`: all four now name `TW_DEV_SERVER_URL` / `TW_BUILD_DIR`.

Took the assertion suggestion past one case: `2b1e09d` pins the full message in every case. Verified it catches the drift rather than assuming it does, by putting the `TW_HOST_` infix back:

```console
--- FAIL: TestConfig_Validate/rejects_invalid_values/dev_server_url_with_a_non-http_scheme
    want err: containing "TW_DEV_SERVER_URL must use scheme..."; got: "TW_HOST_DEV_SERVER_URL must use scheme..."
```

Same commit moves the file onto the CONTRIBUTING.md conventions (`package config`, `want`/`got` over `pkg/require`). Being in-package reaches the unexported `validate()` methods, so `TW_SESSION_*` is covered now: 7 of 13 validation branches to all 13.

server/internal/middleware/session.go
```diff
@@ -0,0 +1,170 @@
...
+					Secure:   opts.Secure,
+					HttpOnly: true,
+					SameSite: http.SameSiteLaxMode,
+				})
+
+				// The save can outlive the request context, which a client that
+				// disconnects mid-response cancels.
+				ctx := context.WithoutCancel(r.Context())
+
+				// The response is already in flight, so the status code cannot
```
YimingIsCOLD: I think this comment is wrong. Status code and body are still able to modify.

YimingIsCOLD: Confirmed. The hook runs before the wrapped writer forwards (`WriteHeader` L32, `Write` L39, `Flush` L46), so nothing has reached the wire and both status and body are still writable. That is why `http.SetCookie` at L118 works at all.

Checked against a real `net/http` server, with the hook rendering a 500 over a handler that had already called `WriteHeader(200)` and written a body:

```
status = 500
body   = "Internal Server Error<h1>handler body</h1>"
```

Overriding *cleanly* would also need the handler's later writes turned into no-ops, or its body trails the error text. We are keeping log-and-return, so the comment now states that as policy rather than a constraint that does not exist.

server/internal/middleware/session.go
YimingIsCOLD: Good catch, fixed. `Commit` runs first and the cookie is written only on success, so a failed commit leaves the client on its existing cookie and entry.

Kept `Commit` before `Drop`. Neither op is transactional and the costs are asymmetric: commit-first failing leaves the user on their pre-login session (recoverable), while drop-first failing at the commit deletes the old entry without writing the new one, destroying the session.

The entry that lingers when `Drop` fails is not a fixation hole. `Commit` writes under the *rotated* ID, so the old ID keeps its pre-login anonymous snapshot. Verified with a `memstore` whose `Drop` always fails: replaying the planted ID returns an anonymous session (`user = <nil>`). The drop is garbage collection, not the defense, so it can stay best effort.

Test gap closed: commit-failure now asserts no cookie, drop-failure asserts the rotated cookie. Mutation-checked by moving `SetCookie` back above `Commit`.

Separately, the `data` map does cross the privilege boundary and rotation does not touch it, so `SetUser` now clears it on the anonymous to authenticated transition.

server/internal/config/config.go
YimingIsCOLD: Fixed in [feb9829](https://github.com/String-dxd/teacher-workspace/pull/59/changes/feb9829f96bcc372ec6becd2fbbfe0d2ad744fb0). 

Using `http.Cookie` to validate the cookie name.

server/internal/config/config.go
YimingIsCOLD: Fixed in https://github.com/String-dxd/teacher-workspace/pull/59/changes/03c524cd718b746bba597d09bd29cdbf589e01b8.

Both TTLs now floor at `1s` with your suggested message. Added a `500ms` case per field to `TestSessionConfig_validate`. `1500ms` still passes and still ships `Max-Age=1`, so cookie and store stay half a second apart. Leaving truncate-vs-round alone

server/internal/middleware/session.go
YimingIsCOLD: Fixed in: https://github.com/String-dxd/teacher-workspace/pull/59/changes/83bb0fce2d3595af8d8dccb06443e5d168565368. 

Added `Cache-Control: no-store` header to session-scoped responses. Skip the `Vary` header for now because `Cache-Control: no-store` leaves nothing stored for it to key.

server/internal/handler/handler.go
YimingIsCOLD: Fixed in: https://github.com/String-dxd/teacher-workspace/pull/59/changes/bceb45458eb51769198c575070e56554a8f37e6f

`TestHandler_Register` drives the mux with a counting stand-in for the session middleware: `/static/...` asserts zero invocations, `/dashboard` asserts one, in both environments

server/internal/middleware/session.go
YimingIsCOLD: Fixed in: https://github.com/String-dxd/teacher-workspace/pull/59/changes/46046fc41ce472e6dcb144c8e3cbf27704ff8174

Deferred executing the save on non-committed status code.

santosral (approved):

PR #69, opened 2026-07-30
feat(`server/proxy`): strip cookie and attach signed JWT on proxied requests

### LINKED ISSUE #44: feat(`server/proxy`): strip cookie and attach signed JWT on proxied requests

santosral:
Closes #44.

## 🚀 Summary

Strips the session cookie from proxied MFE API requests and replaces it with a short-lived signed JWT, so downstream backends can authenticate Teacher Workspace without ever seeing the user's session credential.

Both open questions on the issue are resolved as agreed in the comments: `aud` uses the short code (`pg`, `si`), and each app has its own signing key.

## ✏️ Changes

- Mint a per-request HS256 token in `handler.proxy`, signed with the matched app's key, `aud` set to that app's short code. Claims are `iss`, `aud`, `iat`, `exp`, with no user identity.
- Strip `Cookie` and set `Authorization: Bearer <jwt>` on the outbound request from inside the `Rewrite` hook. `ReverseProxy` deletes every header the client names in `Connection` before `Rewrite` runs, so setting it on the inbound request would let `Connection: Authorization` strip the token.
- Add `TW_API_PROXY_POSTS_SIGNING_KEY` and `TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY`, validated at startup as non-empty and at least 32 bytes (RFC 7518, section 3.2, for HS256)
- Add `TW_API_PROXY_TOKEN_TTL`, default `1m`, floor `1s`. `jwt.NewNumericDate` truncates to whole seconds, so a sub-second TTL mints a token whose `exp` equals its `iat`.
- Leave both keys empty in `Default()` so a missing secret fails startup
- Answer 500 when signing fails, alongside the existing 404 for an unknown app
- Add `golang-jwt/jwt/v5`

## 🧪 Test Plan

`go test -race ./...` and `make lint` clean.

- `TestHandler_proxy/strips the cookie and attaches a signed JWT in Headers` drives `h.proxy` against a live backend per app: no `Cookie` forwarded, a `Bearer` token present (AC1)
- `TestSignToken` covers both apps: HS256, a claim set of exactly `aud`/`exp`/`iat`/`iss`, `iss` of `TW`, the audience given, and `exp - iat` equal to the TTL given
- `TestAPIProxyConfig_validate` gains rows for a missing and a too-short key per app, and zero, negative, and sub-second TTLs. `TestDefault` pins both keys empty and the TTL at `1m`.

Verified by hand against a local posts backend that verifies the token and reports its claims: with a cookie, 200 and the backend sees no `Cookie`; without a cookie, 200 and the same claims, so a cookie-less request is still signed (AC3). The response carries no `Authorization`.


server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,158 @@
...
+	transport.MaxIdleConnsPerHost = opts.MaxIdleConnsPerHost
+
+	loggerFor := logger
+	if loggerFor == nil {
+		loggerFor = func(context.Context) *slog.Logger { return slog.Default() }
+	}
+
+	tokens := signer{audience: target.Audience, key: target.SigningKey, ttl: opts.TokenTTL}
+
+	proxy := &httputil.ReverseProxy{
```
guangshinhaha: **[Medium] The proxy strips cookies on the way out but forwards the downstream's `Set-Cookie` back onto the Teacher Workspace origin**

```go
proxy := &httputil.ReverseProxy{        // ← no ModifyResponse: response headers pass through verbatim
	Transport: transport,
	Rewrite: func(pr *httputil.ProxyRequest) {
		pr.SetURL(target.BaseURL)
		pr.SetXForwarded()
		...
	},
	ErrorHandler: func(w http.ResponseWriter, r *http.Request, err error) { ... },
}
```

**Problem:** The package doc states the point of the design is that "a downstream authenticates Teacher Workspace without ever seeing the end user's session credential" — but the guard is one-directional. Nothing filters the response, and `TestNew/returns the downstream response unchanged` locks in that every `Set-Cookie` the downstream emits is replayed to the browser. Because `/api/{app}/` is served from the Teacher Workspace origin, a downstream that is compromised (or simply buggy) can write cookies scoped to that origin: `Set-Cookie: tw_session=<attacker-chosen>; Path=/api` is stored by the browser alongside the real session cookie, and — being the more specific path — is sent first on every subsequent `/api/*` request, giving the attacker a session-fixation foothold on the very credential this package exists to isolate. It can equally clobber any other first-party cookie (CSRF token, feature flags).

**Suggestion:** Drop response cookies at the same boundary the request cookies are dropped, so the trust boundary is symmetric:

```go
proxy := &httputil.ReverseProxy{
	Transport: transport,
	Rewrite:   func(pr *httputil.ProxyRequest) { ... },
	// A downstream authenticates via the bearer token, so it has no business
	// writing cookies on the Teacher Workspace origin.
	ModifyResponse: func(res *http.Response) error {
		res.Header.Del("Set-Cookie")
		return nil
	},
	ErrorHandler: func(w http.ResponseWriter, r *http.Request, err error) { ... },
}
```

If a downstream genuinely needs to set a cookie, that is worth making an explicit, named allowance rather than the default.

---
*🤖 aif-code-review · claude-opus-5*

server/internal/config/config.go
```diff
@@ -143,6 +219,79 @@ func (c SessionConfig) validate() error {
...
+
+// Targets maps each configured app to its downstream, omitting unset apps.
+func (c APIProxyConfig) Targets() map[string]Target {
+	targets := make(map[string]Target)
+
+	for name, app := range c.apps() {
+		if app.baseURL == nil {
+			continue
+		}
+		targets[name] = Target{BaseURL: app.baseURL, SigningKey: app.signingKey}
```
guangshinhaha: **[Medium] `Targets()` hands out an app with an empty signing key, and nothing downstream of it refuses to sign with one**

```go
func (c APIProxyConfig) Targets() map[string]Target {
	targets := make(map[string]Target)

	for name, app := range c.apps() {
		if app.baseURL == nil {
			continue
		}
		targets[name] = Target{BaseURL: app.baseURL, SigningKey: app.signingKey} // ← key may be "" or 3 bytes
	}

	return targets
}
```

**Problem:** Set `TW_API_PROXY_POSTS_BASE_URL` without `TW_API_PROXY_POSTS_SIGNING_KEY` and `Targets()` returns `{"posts": {BaseURL: …, SigningKey: ""}}`. `handler.Register` mounts that verbatim, and `signer.sign()` calls `jwt.SignedString([]byte(""))` — `jwt/v5`'s HMAC signer has no minimum-key check, so it happily returns a well-formed HS256 token that anybody who guesses "the key is empty" can forge for the `posts` audience. The only thing standing between that and production is `Config.Validate()` being called by a different package in `main`; `apiproxy` and `handler` themselves never verify that the credential they are minting is a credential. That this path is live and silent is demonstrated by this PR's own tests — `TestHandler_Register/returns 502 when the app's downstream is unreachable`, `/merges a base url query into every proxied request` and `/keeps concurrent proxied requests separate` all mount a proxy with `PostsSigningKey` unset and every request is signed with an empty key without a word of complaint.

**Suggestion:** Enforce the invariant where the key is used, so it holds for every caller rather than only for callers that remember to call `Validate()`:

```go
// minKeyLength mirrors the config floor: HMAC-SHA256 with a key shorter than
// its own output produces a signature that is cheap to forge.
const minKeyLength = 32

func (s signer) sign() (string, error) {
	if len(s.key) < minKeyLength {
		return "", fmt.Errorf("signing key must be at least %d bytes; got %d", minKeyLength, len(s.key))
	}

	now := time.Now()
	...
}
```

That already routes through the existing `failed to sign proxy token` branch, so a misconfigured app fails loudly per request instead of forwarding a forgeable token.

---
*🤖 aif-code-review · claude-opus-5*

server/internal/config/config.go
```diff
@@ -143,6 +219,79 @@ func (c SessionConfig) validate() error {
...
+	apps := c.apps()
+	// Sorted so the reported errors do not reorder between runs.
+	for _, name := range slices.Sorted(maps.Keys(apps)) {
+		app := apps[name]
+		if app.baseURL == nil {
+			continue
+		}
+
+		if app.baseURL.Scheme != "http" && app.baseURL.Scheme != "https" {
+			errs = append(errs, fmt.Errorf("%s must use scheme http or https; got %q", app.baseURLVar, app.baseURL))
```
guangshinhaha: **[Medium] A base URL carrying userinfo is accepted, silently discarded, and echoed unredacted into startup error logs**

```go
if app.baseURL.Scheme != "http" && app.baseURL.Scheme != "https" {
	errs = append(errs, fmt.Errorf("%s must use scheme http or https; got %q", app.baseURLVar, app.baseURL)) // ← %q on *url.URL prints user:pass
}
if app.baseURL.Host == "" {
	errs = append(errs, fmt.Errorf("%s must include host[:port]; got %q", app.baseURLVar, app.baseURL))
}
```

**Problem:** Two failures share one root cause — validation treats userinfo as ordinary URL content.

1. `TestAPIProxyConfig_validate/accepts valid base urls/with userinfo` asserts `http://user:pass@pg.example.com` is a valid target. It is not usable: `ProxyRequest.SetURL` copies only `Scheme`, `Host`, `Path` and `RawQuery` (`rewriteRequestURL` in `net/http/httputil`), so `URL.User` never reaches the outbound request; and even if it did, `apiproxy` overwrites `Authorization` with the bearer token. An operator who configures `TW_API_PROXY_POSTS_BASE_URL=https://svc:hunter2@posts.internal` gets a proxy that starts cleanly and then 401s against the downstream forever, with nothing in the logs pointing at the dropped credential.
2. `%q` on a `*url.URL` invokes `String()`, which does **not** redact the password. So `TW_API_PROXY_POSTS_BASE_URL=ftp://svc:hunter2@posts.internal` writes `hunter2` into the startup error — the exact leak `apiproxy.logProxy` deliberately avoids by calling `upstream.Redacted()`.

**Suggestion:** Reject userinfo outright and redact it everywhere it is printed:

```go
if app.baseURL.User != nil {
	// SetURL drops userinfo and the proxy overwrites Authorization, so a
	// credential here is silently ignored rather than applied.
	errs = append(errs, fmt.Errorf("%s must not contain userinfo; got %q", app.baseURLVar, app.baseURL.Redacted()))
}
if app.baseURL.Scheme != "http" && app.baseURL.Scheme != "https" {
	errs = append(errs, fmt.Errorf("%s must use scheme http or https; got %q", app.baseURLVar, app.baseURL.Redacted()))
}
if app.baseURL.Host == "" {
	errs = append(errs, fmt.Errorf("%s must include host[:port]; got %q", app.baseURLVar, app.baseURL.Redacted()))
}
```

and flip the `with userinfo` case in `TestAPIProxyConfig_validate` from the accepted table to the rejected one.

---
*🤖 aif-code-review · claude-opus-5*

server/internal/config/config.go
```diff
@@ -143,6 +219,79 @@ func (c SessionConfig) validate() error {
...
+		}
+		if app.baseURL.Host == "" {
+			errs = append(errs, fmt.Errorf("%s must include host[:port]; got %q", app.baseURLVar, app.baseURL))
+		}
+
+		// A routed app without a key would forward unsigned, so fail startup
+		// rather than at request time.
+		switch {
+		case app.signingKey == "":
```
guangshinhaha: **[Low] The "distinct per app" key requirement is documented but never enforced**

```go
// A routed app without a key would forward unsigned, so fail startup
// rather than at request time.
switch {
case app.signingKey == "":                                  // ← per-app checks only; no cross-app comparison
	errs = append(errs, fmt.Errorf("%s is required when %s is set", app.signingKeyVar, app.baseURLVar))
case len(app.signingKey) < minSigningKeyLength:
	errs = append(errs, fmt.Errorf("%s must be at least %d bytes; got %d", app.signingKeyVar, minSigningKeyLength, len(app.signingKey)))
}
```

**Problem:** `.env.example` states the invariant plainly — keys must be "distinct per app so one leaked key cannot forge calls to the others" — but `validate()` only ever looks at one app at a time. Today `apps()` has a single entry so nothing can violate it; the moment a second app lands (the tests already rehearse `student-insight`), copy-pasting the same 32-byte string into both `TW_API_PROXY_*_SIGNING_KEY` vars starts the server without complaint and collapses the per-audience isolation that `TestNew_Signing/gives each app an audience only its own key verifies` exists to guarantee. The loop is already the right shape to catch it; it just does not.

**Suggestion:** Enforce it while the code has exactly one app, so onboarding the second cannot regress it:

```go
seen := make(map[string]string, len(apps))
for _, name := range slices.Sorted(maps.Keys(apps)) {
	app := apps[name]
	...
	if app.signingKey != "" {
		if other, dup := seen[app.signingKey]; dup {
			errs = append(errs, fmt.Errorf("%s must differ from %s; one leaked key would forge calls to both", app.signingKeyVar, other))
		} else {
			seen[app.signingKey] = app.signingKeyVar
		}
	}
}
```

---
*🤖 aif-code-review · claude-opus-5*

server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,158 @@
...
+
+func logProxy(logger *slog.Logger, r *http.Request, level slog.Level, msg string, extra ...slog.Attr) {
+	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")
+	upstream := *r.URL
+	upstream.RawQuery = ""
+
+	logger.LogAttrs(r.Context(), level, msg, append([]slog.Attr{
+		slog.String("method", r.Method),
+		slog.String("path", inboundPath),
+		slog.String("upstream", upstream.Redacted()),
```
guangshinhaha: **[Low] `upstream` in the log record is the inbound path, not an upstream, on two of the three call sites**

```go
func logProxy(logger *slog.Logger, r *http.Request, level slog.Level, msg string, extra ...slog.Attr) {
	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")
	upstream := *r.URL
	upstream.RawQuery = ""

	logger.LogAttrs(r.Context(), level, msg, append([]slog.Attr{
		slog.String("method", r.Method),
		slog.String("path", inboundPath),
		slog.String("upstream", upstream.Redacted()), // ← only a downstream URL when called from ErrorHandler
	}, extra...)...)
}
```

**Problem:** `ErrorHandler` receives the outbound request, whose `URL` `Rewrite` already pointed at the downstream — that is the case the field is named for, and `TestNew/names the method and upstream in the gateway error log` pins it. The other two call sites (`refused a streaming request`, `failed to sign proxy token`) pass the *inbound* `r`, whose `URL` is the server-side relative form. A refused SSE request to `/api/posts/events` therefore logs `"path":"/api/posts/events","upstream":"/api/posts/events"` — the same value twice, one of them under a key that tells an on-call engineer they are looking at the downstream address. Neither of those two messages has a test, so the duplication is invisible.

**Suggestion:** Only emit `upstream` when it really is one:

```go
func logProxy(logger *slog.Logger, r *http.Request, level slog.Level, msg string, extra ...slog.Attr) {
	inboundPath, _, _ := strings.Cut(r.RequestURI, "?")

	attrs := []slog.Attr{
		slog.String("method", r.Method),
		slog.String("path", inboundPath),
	}
	// r is the outbound request only when the proxy itself reports the error;
	// before that r.URL is still the inbound, relative one.
	if r.URL.Host != "" {
		upstream := *r.URL
		upstream.RawQuery = ""
		attrs = append(attrs, slog.String("upstream", upstream.Redacted()))
	}

	logger.LogAttrs(r.Context(), level, msg, append(attrs, extra...)...)
}
```

---
*🤖 aif-code-review · claude-opus-5*

server/internal/apiproxy/apiproxy.go
```diff
@@ -0,0 +1,158 @@
...
+			logProxy(loggerFor(r.Context()), r, slog.LevelWarn, "refused a streaming request")
+			http.Error(w, "Not Implemented", http.StatusNotImplemented)
+
+			return
+		}
+
+		token, err := tokens.sign()
+		if err != nil {
+			logProxy(loggerFor(r.Context()), r, slog.LevelError, "failed to sign proxy token", slog.Any("err", err))
+			http.Error(w, "Bad Gateway", http.StatusBadGateway)
```
guangshinhaha: **[Low] A local signing failure is reported to the caller as 502 Bad Gateway**

```go
token, err := tokens.sign()
if err != nil {
	logProxy(loggerFor(r.Context()), r, slog.LevelError, "failed to sign proxy token", slog.Any("err", err))
	http.Error(w, "Bad Gateway", http.StatusBadGateway) // ← the gateway was never contacted
	return
}
```

**Problem:** `sign()` fails for reasons entirely internal to this server — an unusable key, `crypto/sha256` unavailable — and at that point no connection to the downstream has been attempted. Returning 502 tells the client, and every dashboard built on status codes, that the downstream is unhealthy. During an incident that sends whoever is on call to inspect a `posts` service that is working fine, while the real fault (a bad `TW_API_PROXY_*_SIGNING_KEY`) sits one line away in the logs. It also makes the 502 emitted here indistinguishable from the genuine gateway failures raised by `ErrorHandler`.

**Suggestion:** Report the fault to the layer that owns it:

```go
token, err := tokens.sign()
if err != nil {
	logProxy(loggerFor(r.Context()), r, slog.LevelError, "failed to sign proxy token", slog.Any("err", err))
	// The downstream was never reached: this is our fault, not the gateway's.
	http.Error(w, "Internal Server Error", http.StatusInternalServerError)

	return
}
```

---
*🤖 aif-code-review · claude-opus-5*

guangshinhaha:
## Code Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High     | 0 |
| Medium   | 3 |
| Low      | 3 |

## What Looks Good
- Minting a fresh per-request token with only `iss`/`aud`/`iat`/`exp` and asserting that claim set exactly (`carries no claim beyond iss, aud, iat and exp`) is the right call — it makes "no user identity travels downstream" a property the test suite enforces rather than a comment.
- Keying the audience off the route segment and giving each app its own signing key, then proving cross-app forgery fails with `jwt.ErrSignatureInvalid`, means a leaked `posts` key cannot be replayed against a future app.
- Validating `TW_API_PROXY_RESPONSE_HEADER_TIMEOUT` and `TW_API_PROXY_REQUEST_TIMEOUT` against `TW_SERVER_WRITE_TIMEOUT` at startup catches the misconfiguration where the server kills the connection before the proxy can emit its own 502 — a failure that would otherwise only show up under load.
- `clientGone` distinguishing a caller hang-up from a genuine downstream failure, with `TestClientGone` covering the wrapped `*url.Error` and deadline-vs-cancel cases, keeps the error log signal-bearing instead of noisy.
- Registering `/api/` as an explicit `NotFoundHandler` so an unconfigured app falls through to 404 rather than the SPA index, with tests in both development and production modes, closes a genuinely easy-to-miss routing hole.

---
*🤖 aif-code-review · claude-opus-5*

guangshinhaha:
@santosral — code review done, findings below. Each one is also posted as an inline comment on the relevant line.

**Severity counts:** Critical 0 · High 0 · Medium 3 · Low 3

### Medium

**1. `Set-Cookie` from a downstream is replayed onto the Teacher Workspace origin** — `server/internal/apiproxy/apiproxy.go:57`
The proxy strips `Cookie` on the way out but has no `ModifyResponse`, so a downstream's `Set-Cookie` flows straight back to the browser on our origin. A compromised or misbehaving downstream can write or fixate cookies (including `tw_session; Path=/api`) on exactly the origin this package exists to protect.

**2. An empty/short `SigningKey` reaches `signer.sign()` and produces a forgeable token** — `server/internal/config/config.go:283`
`Targets()` will happily return an app with an empty signing key; `handler.Register` mounts it and `sign()` HMACs with `[]byte("")` (jwt/v5 enforces no minimum key length). Only the cross-package `Config.Validate()` call in `main` prevents this today — and three of this PR's own handler tests run proxies with an empty key without complaint.

**3. Base-URL userinfo is silently dropped, and the password is logged unredacted** — `server/internal/config/config.go:234`
The test table explicitly accepts a base URL containing userinfo, but `rewriteRequestURL` never copies `URL.User` (and `Authorization` is overwritten regardless), so the credential vanishes without a word. Separately, `%q` on a `*url.URL` prints the password in the clear into startup error logs.

### Low

**4. No cross-app signing-key uniqueness check** — `server/internal/config/config.go:243`
`.env.example` documents that signing keys must be "distinct per app", but `validate()` only runs per-app checks. Unreachable with one app configured; regresses the moment a second one lands.

**5. `logProxy`'s `upstream` attribute isn't always an upstream** — `server/internal/apiproxy/apiproxy.go:149`
It only holds a downstream URL when called from `ErrorHandler`. The streaming-refusal and sign-failure call sites pass the inbound request, so `path` and `upstream` log as the same relative path.

**6. Local signing failure returns 502 Bad Gateway** — `server/internal/apiproxy/apiproxy.go:86`
No connection to the downstream was attempted, so 502 points on-call at a healthy service and makes this indistinguishable from a real gateway failure.

### What looks good

- Per-request minimal-claim tokens, with a test asserting the exact claim set rather than just the happy path.
- Per-app audience **and** per-app key, backed by a cross-forgery test proving app A's token can't be used against app B.
- Proxy-timeout-vs-`WriteTimeout` validation at startup, so a misconfiguration fails fast instead of surfacing as truncated responses in production.
- `clientGone` correctly discriminating a client hang-up from a genuine upstream failure — including `DeadlineExceeded` landing on the right side.
- The explicit `/api/` `NotFoundHandler`, so unmatched API paths don't fall through to the SPA handler.

---
*🤖 aif-code-review · claude-opus-5*

server/internal/apiproxy/apiproxy.go
santosral: Thanks for the review. Flagging that this branch has been rebased onto `main` and the approach reworked, so these comments now point at code that no longer exists.

The `server/internal/apiproxy` package is gone. Signing lives in `server/internal/handler` alongside the proxy handler it serves, since the claim set is Teacher Workspace policy rather than a reusable helper. `Targets()` is gone too, replaced by the existing app switch in `handler.proxy`.

Resolving all six threads on that basis rather than leaving them against deleted files. Two of the underlying concerns are addressed in the new shape: both signing keys are now validated at startup (non-empty, at least 32 bytes) so a missing one fails the boot rather than being handed out empty, and a signing failure answers 500 rather than 502. The downstream `Set-Cookie` point stands and is called out as a follow-up in the PR description, since #44 puts response headers out of scope.

Re-review welcome once the new commits are pushed.

.env.example
```diff
@@ -17,4 +17,7 @@ TW_SERVER_PORT=3000
 
 # API proxy configuration
 # TW_API_PROXY_STUDENT_INSIGHTS_BASE_URL=http://127.0.0.1:3002
+# TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY=nTfUafx3rvFlGuCiDUuhWUnx7+DRmrQyBBAH0cNMZzc=
```
santosral: Change the values of the signing keys to "secret" to avoid CI flagging this as a security risk.

server/internal/config/config.go
```diff
@@ -78,6 +82,7 @@ func Default() Config {
 		APIProxy: APIProxyConfig{
 			StudentInsightsBaseURL: must(url.Parse("http://127.0.0.1:3002")),
 			PostsBaseURL:           must(url.Parse("http://127.0.0.1:3003")),
+			TokenTTL:               1 * time.Minute,
```
santosral: Add the signing keys to make the config valid.

server/internal/config/config.go
```diff
@@ -171,6 +176,13 @@ func (c APIProxyConfig) validate() error {
 			errs = append(errs, fmt.Errorf("TW_API_PROXY_STUDENT_INSIGHTS_BASE_URL must include host[:port]; got %q", c.StudentInsightsBaseURL))
 		}
 	}
+	if c.StudentInsightsSigningKey == "" {
+		errs = append(errs, errors.New("TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY is required"))
+	} else {
+		if siKeyLength := len(c.StudentInsightsSigningKey); siKeyLength < sha256.Size {
+			errs = append(errs, fmt.Errorf("TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY must be at least %d bytes; got %d", sha256.Size, siKeyLength))
+		}
```
santosral: Can the signing key can be more than 32 bytes?

server/internal/config/config_test.go
```diff
@@ -328,8 +339,11 @@ func TestAPIProxyConfig_validate(t *testing.T) {
 		for _, scheme := range []string{"http", "https"} {
 			t.Run(scheme, func(t *testing.T) {
 				cfg := APIProxyConfig{
-					StudentInsightsBaseURL: &url.URL{Scheme: scheme, Host: "student-insights.example.com"},
-					PostsBaseURL:           &url.URL{Scheme: scheme, Host: "posts.example.com"},
+					StudentInsightsBaseURL:    &url.URL{Scheme: scheme, Host: "student-insights.example.com"},
+					StudentInsightsSigningKey: "0123456789abcdef0123456789abcdef",
```
santosral: Use `secret`

server/internal/handler/handler.go
```diff
@@ -24,15 +24,11 @@ func New(cfg *config.Config) *Handler {
 	h := &Handler{
 		cfg: cfg,
 		studentInsightsProxy: &stdhttputil.ReverseProxy{
-			Rewrite: func(pr *stdhttputil.ProxyRequest) {
-				pr.SetURL(cfg.APIProxy.StudentInsightsBaseURL)
-			},
+			Rewrite:      proxyRewriter(cfg.APIProxy.StudentInsightsBaseURL),
```
santosral: Remove proxyRewriter to inline inbound request transformation in Handler.

server/internal/handler/proxy_test.go
```diff
@@ -138,6 +142,129 @@ func TestHandler_proxy(t *testing.T) {
 		}
 	})
 
+	t.Run("strips the cookie and attaches a signed JWT in Headers", func(t *testing.T) {
```
santosral: split cookie removal from JWT attachment.

server/internal/config/config.go
```diff
@@ -171,6 +178,13 @@ func (c APIProxyConfig) validate() error {
 			errs = append(errs, fmt.Errorf("TW_API_PROXY_STUDENT_INSIGHTS_BASE_URL must include host[:port]; got %q", c.StudentInsightsBaseURL))
 		}
 	}
+	if c.StudentInsightsSigningKey == "" {
+		errs = append(errs, errors.New("TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY is required"))
+	} else {
+		if siKeyLength := len(c.StudentInsightsSigningKey); siKeyLength < sha256.Size {
```
```suggestion
		if keyLen := len(c.StudentInsightsSigningKey); keyLen < sha256.Size {
```

Add a comment to RFC 7518 on why we need at least 32 bytes key length.

server/internal/config/config.go
```diff
@@ -76,8 +80,11 @@ func Default() Config {
 			AuthenticatedTTL: 30 * time.Minute,
 		},
 		APIProxy: APIProxyConfig{
-			StudentInsightsBaseURL: must(url.Parse("http://127.0.0.1:3002")),
-			PostsBaseURL:           must(url.Parse("http://127.0.0.1:3003")),
+			StudentInsightsBaseURL:    must(url.Parse("http://127.0.0.1:3002")),
+			StudentInsightsSigningKey: "secret",
```
```suggestion
			StudentInsightsSigningKey: "0123456789abcdef0123456789abcdef",
```

The signing key need to be at least 32 bytes long.

server/internal/config/config_test.go
```diff
@@ -65,6 +68,8 @@ func TestConfig_Validate(t *testing.T) {
 		for _, scheme := range []string{"http", "https"} {
 			t.Run(scheme, func(t *testing.T) {
 				cfg := Default()
+				cfg.APIProxy.StudentInsightsSigningKey = "0123456789abcdef0123456789abcdef"
```
YimingIsCOLD: Remove these two line if not necessary.

server/internal/config/config_test.go
```diff
@@ -375,9 +389,46 @@ func TestAPIProxyConfig_validate(t *testing.T) {
...
 				want:   `TW_API_PROXY_POSTS_BASE_URL must include host[:port]; got "http:"`,
 			},
+			{
+				name:   "missing student insights signing key",
+				mutate: func(c *APIProxyConfig) { c.StudentInsightsSigningKey = "" },
+				want:   "TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY is required",
+			},
+			{
+				name:   "short student insights signing key",
+				mutate: func(c *APIProxyConfig) { c.StudentInsightsSigningKey = "0123456789abcdef0123456789abcde" },
```
YimingIsCOLD:
```suggestion
				mutate: func(c *APIProxyConfig) { c.StudentInsightsSigningKey = "secret" },
```

server/internal/config/config_test.go
```diff
@@ -375,9 +389,46 @@ func TestAPIProxyConfig_validate(t *testing.T) {
...
+			},
+			{
+				name:   "sub-second token TTL",
+				mutate: func(c *APIProxyConfig) { c.TokenTTL = 500 * time.Millisecond },
+				want:   "TW_API_PROXY_TOKEN_TTL must be at least 1s; got 500ms",
+			},
 		} {
 			t.Run(tt.name, func(t *testing.T) {
 				cfg := Default().APIProxy
+				cfg.StudentInsightsSigningKey = "0123456789abcdef0123456789abcdef"
```
YimingIsCOLD: Remove these two line if not necessary.

server/internal/handler/proxy.go
```diff
@@ -4,27 +4,54 @@ import (
...
 	}
 
+	now := time.Now()
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.RegisteredClaims{
+		Issuer:    "TW",
+		Audience:  jwt.ClaimStrings{aud},
+		IssuedAt:  jwt.NewNumericDate(now),
+		ExpiresAt: jwt.NewNumericDate(now.Add(h.cfg.APIProxy.TokenTTL)),
+	})
+	signedJWT, err := token.SignedString([]byte(signingKey))
```
YimingIsCOLD:
```suggestion
	signedToken, err := token.SignedString([]byte(signingKey))
```

server/internal/handler/proxy_test.go
```diff
@@ -138,33 +136,7 @@ func TestHandler_proxy(t *testing.T) {
 		}
 	})
 
-}
-
-func TestProxyErrorHandler(t *testing.T) {
```
YimingIsCOLD, line 143: Why remove `TestProxyErrorHandler`?

.env.example
```diff
@@ -17,4 +17,7 @@ TW_SERVER_PORT=3000
 
 # API proxy configuration
 # TW_API_PROXY_STUDENT_INSIGHTS_BASE_URL=http://127.0.0.1:3002
+# TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY=secret
```
YimingIsCOLD:
```suggestion
# TW_API_PROXY_STUDENT_INSIGHTS_SIGNING_KEY=0123456789abcdef0123456789abcdef
```

server/internal/handler/handler_test.go
```diff
@@ -155,4 +231,528 @@ func TestHandler_Register(t *testing.T) {
...
+		if got := rec.Body.String(); strings.Contains(got, "proxied:") {
+			t.Errorf("want: not containing %q; got: %q", "proxied:", got)
+		}
+	})
+
+	t.Run("returns 502 when the app's downstream is unreachable", func(t *testing.T) {
+		discardLogs(t)
+
+		cfg := config.Default()
+		cfg.APIProxy.Posts = deadDownstream(t)
```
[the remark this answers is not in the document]
santosral: Rebased and changed the approached. These reviews comments are stale.

.env.example
santosral: Related to https://github.com/String-dxd/teacher-workspace/pull/69#discussion_r3765428154 and will update the sample keys

server/internal/config/config.go
santosral: Updated by commit: e1e57bc98b9431063544d8548779f8c246b587ac

server/internal/config/config.go
santosral: Related to https://github.com/String-dxd/teacher-workspace/pull/69#discussion_r3765428154

server/internal/config/config_test.go
santosral: Superseded by https://github.com/String-dxd/teacher-workspace/pull/69#discussion_r3765428154

server/internal/handler/handler.go
santosral: Not remove because we need to use the `setUrl() `from the `ProxyRequest` to forward it to the targeted backends.

server/internal/handler/proxy_test.go
santosral: Resolved in 2f320b1178d1ac6a3f407b519fa7102940cfcdc0

server/internal/config/config.go
santosral: resolved a454dff3b09d6268fd6c807a40007bbb3df7aaec

server/internal/config/config.go
santosral: I've used `"a-string-secret-at-least-256-bits-long"` for readability. Changed on a454dff3b09d6268fd6c807a40007bbb3df7aaec

server/internal/config/config_test.go
santosral: Removed on ab1883d34e6697b5eea9b00a34092c1eb2d4b0d0

server/internal/config/config_test.go
santosral: Stale from our last discussion where we need it to be above 32 bytes for `HS256`.

server/internal/config/config_test.go
santosral: Removed on ab1883d34e6697b5eea9b00a34092c1eb2d4b0d0

server/internal/handler/proxy.go
santosral: I've used JWT because the function comments says it returns signed JWT. 

Renamed to `signedToken` 2993a6050d1e525d24729cf6374463bd6b670dfa

server/internal/handler/proxy_test.go
santosral: Reverted it back. 

I was expereminting to refactor the proxy handler function and pushed this changes.

.env.example
santosral: Done and used human readable secret instead 2993a6050d1e525d24729cf6374463bd6b670dfa

server/internal/config/config.go
```diff
@@ -158,6 +164,10 @@ func (c SessionConfig) validate() error {
 	return errors.Join(errs...)
 }
 
+// RFC 7518, section 3.2 requires a key at
+// least as big as hash output (HS256).
+const minSigningKeyBytes = 32
```
YimingIsCOLD: No need to have a const. I think we can use `sha256.Size` because RFC 7518 mention it should be more than or equal to the output length. The comment that mentioned RFC 7518 can be above the if statement.

server/internal/handler/proxy_test.go
```diff
@@ -138,6 +138,206 @@ func TestHandler_proxy(t *testing.T) {
 		}
 	})
 
+	t.Run("logs the failure when the backend is unreachable", func(t *testing.T) {
```
YimingIsCOLD: We already have this test in `TestProxyErrorHandler`, why repeat again?

server/internal/handler/proxy_test.go
```diff
@@ -138,6 +138,206 @@ func TestHandler_proxy(t *testing.T) {
...
+		// The rewritten backend URL, not the one the client asked for.
+		if want, got := backendURL.String()+"/hello", entry.Backend; want != got {
+			t.Errorf("want: %q; got: %q", want, got)
+		}
+		if want, got := "connection refused", entry.Err; !strings.Contains(got, want) {
+			t.Errorf("want: containing %q; got: %q", want, got)
+		}
+	})
+
+	t.Run("keeps the query string out of the logs", func(t *testing.T) {
```
YimingIsCOLD: We already have this test in `TestProxyErrorHandler`, why repeat again?

server/internal/handler/proxy_test.go
```diff
@@ -138,6 +138,206 @@ func TestHandler_proxy(t *testing.T) {
...
+		}
+
+		cfg := config.Default()
+		cfg.APIProxy.PostsBaseURL = postsBackendURL
+
+		h := New(&cfg)
+
+		req := httptest.NewRequest(http.MethodGet, "/api/posts/hello", nil)
+		req.SetPathValue("app", "posts")
+		req.AddCookie(&http.Cookie{Name: cfg.Session.Name, Value: "session-value"})
```
YimingIsCOLD: Try not to reference other variable. Replace `cfg.Session.Name` with a hardcoded string.

.env.example
```diff
@@ -18,3 +18,7 @@ TW_SERVER_PORT=3000
 # API proxy configuration
 # TW_API_PROXY_STUDENT_INSIGHTS_BASE_URL=http://127.0.0.1:3002
 # TW_API_PROXY_POSTS_BASE_URL=http://127.0.0.1:3003
+# Generate with: openssl rand -base64 32
```
YimingIsCOLD: I would suggest we remove this comment first. Wait till we know more about how local proxy configured/works then we add it in CONTRIBUTING.md/Development Setup

server/internal/handler/proxy_test.go
```diff
@@ -138,6 +138,206 @@ func TestHandler_proxy(t *testing.T) {
...
+		cfg.APIProxy.StudentInsightsBaseURL = studentInsightsBackendURL
+
+		h := New(&cfg)
+
+		tests := []struct {
+			name string
+			app  string
+		}{
+			{name: "posts", app: "posts"},
+			{name: "student insights", app: "student-insights"},
```
YimingIsCOLD: Right now changing the `app` field to both `posts` or `student-insights` does not break the test, the question we would need ask ourselves is: why do we need to do table driven tests?

I think what we want to test here is whether with different inputs like `posts` or `student-insights`, the JWT is signed correct, place in the correct header and the content in the JWT is valid.

```suggestion
            {name: "posts", app: "posts", signingKey: cfg.APIProxy.PostsSigningKey, wantAud: "pg"},
            {name: "student insights", app: "student-insights", signingKey: cfg.APIProxy.StudentInsightsSigningKey, wantAud: "si"},
```

I think we can add more wants to validate all claims are present.

server/internal/handler/proxy_test.go
YimingIsCOLD: Seems like there are still repeating tests which already tested in the `TestProxyErrorHandler`. I left another comment to address those.

YimingIsCOLD:
CI job just got merged, please rebase to make sure CI job is successful for your branch.

server/internal/config/config.go
santosral: Importing the crypto package just to use the SHA256.size feels wrong to me because we're only using its constant. Let me know if this makes sense or we just import and use the package.

server/internal/handler/proxy_test.go
santosral: Removed 78e82fa7abc2063bd2a1b65bd89ab72f6b51ed75

server/internal/handler/proxy_test.go
santosral: Removed 78e82fa7abc2063bd2a1b65bd89ab72f6b51ed75

server/internal/handler/proxy_test.go
santosral: Replaced 91e037809b4c92067fde150beadc6f3d66aea792

.env.example
santosral: Removed b907b8580d0ab44fbc7bdd9a2988291575354ad5

server/internal/handler/proxy_test.go
santosral: Updated 91e037809b4c92067fde150beadc6f3d66aea792

santosral: Removed the table driven test, the main goal of the test is just to check if the Authorization Header is sent and the JWT have the proper Claims.

santosral: Put back the table driven test to and handle the forward JWT flow, signature, and claims.

server/internal/config/config.go
YimingIsCOLD: I think it is fine. We are using what the stdlib is providing instead of defining ourselves.

YimingIsCOLD (dismissed):
Although the proxy tests still feels weird, we can merge first.

server/internal/config/config.go
santosral: Done

YimingIsCOLD (approved):

PR #71, opened 2026-08-05
chore(`Dockerfile`): add multi-stage image build and CI build job

### LINKED ISSUE #65: chore(`Dockerfile`): add multi-stage image build and CI build job
## User story

As a platform engineer, I want the Go server and the `apps/host` bundle built into a single container image, so that I can run Teacher Workspace anywhere a container runs instead of reproducing the Go, Node, and pnpm toolchains on every machine by hand.

## Background

The only documented way to run Teacher Workspace in production mode is `pnpm build` followed by `TW_ENV=production go run ./server/cmd/tw`. That needs Go 1.26.5, Node 24, and pnpm 11 on the host, and produces no artifact to hand to a deployment target.

The server already supports being packaged: in `production`, `handler.New` serves static files from `cfg.BuildDir` with an `index.html` fallback, and `dotenv.Load` tolerates a missing `.env` so a container can be configured entirely through `TW_*` variables. This issue packages the production path that already exists.

Two deliberate limits on this first pass: **arm64 Alpine only** (the repository is public, so `ubuntu-24.04-arm` runners build it natively with no QEMU), and **no push** (`docker/build-push-action` with `push: false`, so the wiring exists without introducing registry credentials or a tagging scheme).

### Image serves the host app over HTTP

- **Given** a checkout of the repository at a commit where `pnpm build` and `go build ./server/cmd/tw` both succeed
- **When** an image is built from the repository-root `Dockerfile` and a container is started from it with the server port published
- **Then** a `GET /` request returns `200` with the host app's `index.html`, a request for a hashed asset under `/static/` returns `200` with that asset, and a request for an unknown client-side route such as `/anything` returns `200` with `index.html`

### Server starts from environment variables alone

- **Given** the built image, which contains no `.env` file
- **When** a container is started with only `TW_*` environment variables supplied (including a non-default `TW_SERVER_PORT`)
- **Then** the server starts without a configuration error and listens on the port given by `TW_SERVER_PORT`

### Runtime image contains no build toolchain

- **Given** the built image
- **When** its filesystem is inspected
- **Then** it contains the compiled server binary, the host build output, and their runtime dependencies, and it contains no Go toolchain, Node runtime, pnpm, `node_modules`, or repository source tree

### Container shuts down cleanly on SIGTERM

- **Given** a running container started from the image
- **When** the container runtime sends `SIGTERM` (for example via `docker stop`)
- **Then** the server runs its existing graceful shutdown and the container exits with code `0` before the stop timeout elapses, with no orphaned process left behind

### CI fails when the image build breaks

- **Given** a pull request that breaks the image build (for example a type error in `apps/host` or a compile error under `server/`)
- **When** CI runs on that pull request
- **Then** the Docker job fails and its logs identify the build stage that failed, and no image is pushed to any registry

### CI reports the image build on every pull request

- **Given** a pull request that does not break the build
- **When** CI runs on that pull request
- **Then** the Docker job builds the image for `linux/arm64` and reports success as a required-visible check alongside the existing Format, Lint, Go Lint, and Go Test jobs

## Out of scope

- **Docker-based local development.** No `docker-compose.yml` and no development-mode image. The documented `pnpm dev` plus `go run ./server/cmd/tw` flow is unchanged, and the Rsbuild dev-server proxy path stays outside the container.
- **Deployment.** No Cloud Run, Kubernetes, or other environment manifests, and no deploy pipeline. This issue ends at a built image and a green CI job.
- **Registry publishing.** `push` stays `false`. No registry credentials, no `packages: write` permission, no image tagging or versioning scheme.
- **Multi-architecture images.** `linux/arm64` only. `linux/amd64` and multi-platform manifests are a follow-up.
- **A dedicated health endpoint.** No new `/healthz` route is added to the server. Any container health check uses the existing `/` route.
- **Container smoke tests in CI.** The Docker job passes on a successful build. It does not start the container or assert HTTP behaviour, so the runtime scenarios above are verified manually by the implementer rather than by CI.
- **Build-cache tuning in CI.** Registry-backed or GitHub Actions cache export for Docker layers is a follow-up; a correct build takes priority over a fast one.

## Design assets

No Figma or screenshots: this change has no UI surface. The intended stage topology:

```mermaid
flowchart LR
  subgraph build["Build stages (discarded from the final image)"]
    direction TB
    web["host-build<br/>node:24-alpine<br/>pnpm install --frozen-lockfile<br/>pnpm build"]
    gob["server-build<br/>golang:1.26.5-alpine<br/>go build ./server/cmd/tw"]
  end

  subgraph run["Runtime stage (shipped)"]
    final["alpine, linux/arm64<br/>non-root user<br/>TW_ENV=production<br/>listens on TW_SERVER_PORT"]
  end

  web -->|apps/host/dist| final
  gob -->|tw binary| final
```

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

kellylimmm:
## 🚀 Summary

The only documented way to run Teacher Workspace in production mode is `pnpm build` followed by `TW_ENV=production go run ./server/cmd/tw`, which needs Go, Node, and pnpm installed on the host and produces no artifact to hand to a deployment target. This adds a multi-stage `Dockerfile` that packages the Go server and the `apps/host` bundle into a single container image, plus a CI job that builds it on every pull request, so Teacher Workspace can run anywhere a container runs.

This first pass is deliberately limited to **arm64 Alpine only** (the repo is public, so `ubuntu-24.04-arm` runners build it natively with no QEMU) and **no push** (`push: false`, so no registry credentials or tagging scheme are introduced yet).

Closes #65.

## ✏️ Changes

- Added a repository-root `Dockerfile` with three stages: `host-build` (Rsbuild build of `apps/host`), `server-build` (static Go build of `server/cmd/tw`), and a minimal Alpine runtime stage running as a non-root user (`zero`).
- Added `.dockerignore` to keep `node_modules`, `dist`, `.git`, `.env`, and editor directories out of the build context.
- Added a `build-image` job to `.github/workflows/ci.yml`, running alongside the existing Format, Lint, Go Lint, and Go Test jobs with no `needs:` gate.

### Updates from review

- Dropped `TARGETARCH`/`GOARCH`/`GOOS` from the Go build: the base image for that stage already matches the target platform and is always Linux, so `go build` defaults to both correctly without being told.
- Switched dependency fetching to `pnpm fetch`, which only reads `pnpm-lock.yaml`, so adding a new app under `apps/*` later needs no Dockerfile change to copy its manifest in.
- Renamed the container user from `tw` to `zero`, a more generic name.
- Removed unused `.worktrees`/`docs` entries from `.dockerignore` (this repo has neither).
- Held off documenting the container image build in `CONTRIBUTING.md` until the deployment flow is settled.
- Trimmed a few comments that just restated context covered elsewhere, including the one justifying the now-removed `GOOS`/`GOARCH`; collapsed the Go build `RUN` back onto a single line now that it's short enough.
- Considered switching to the `ghcr.io/pnpm/pnpm` base image and a `pnpm`-standalone-binary install instead of `npm install --global pnpm`; kept `npm install` for both: the standalone binary segfaults on arm64 Alpine for every pnpm 11.x release (which this repo's `engines.pnpm: ^11.0.0` requires), and the base-image swap wasn't adopted for this pass.

## 🧪 Test Plan

Built and ran the image locally end to end (Docker daemon started explicitly for this since it wasn't running by default):

- `docker build -t teacher-workspace:test .` succeeds; image is `linux/arm64`, ~7.5 MB
- `GET /` → `200` with `index.html`; a real hashed asset under `/static/` → `200`; `GET /anything` (unknown client route) → `200`, byte-identical to `/`; missing asset under `/static/` → `404`
- Server starts from environment variables alone (no `.env` in the image); `-e TW_SERVER_PORT=8080` is honored and the startup log reflects it
- Runtime image contains no build toolchain: `go`, `node`, `pnpm` all absent; `/app` holds only `tw` and `dist`; no `/src`; no `.env` or `node_modules` anywhere in the filesystem
- Runs as non-root (`uid=100(zero) gid=101(zero)`)
- `docker stop` exits cleanly with code `0` in ~216ms (graceful `SIGTERM` shutdown via exec-form `CMD`)
- Verified version pins are in sync: `golang:1.26.5-alpine3.23` matches `go.mod`'s `go 1.26.5`, and `PNPM_VERSION=11.11.0` matches `pnpm/action-setup`'s `version: 11.11.0` in `ci.yml`
- Verified the two new/changed GitHub Actions (`docker/setup-buildx-action`, `docker/build-push-action`) are pinned to commit SHAs that resolve to their claimed version tags
- Re-verified after the review-feedback changes: `docker build --no-cache` succeeds, container starts, `GET /` → `200`, and `id` inside the container confirms `uid=100(zero) gid=101(zero)`

`build-image` ran for real on this PR and passed, alongside the existing Format, Lint, Go Lint, and Go Test jobs:

| Check | Result | Time |
| --- | --- | --- |
| Build image | ✅ pass | 37s |
| Format | ✅ pass | 26s |
| Go Lint | ✅ pass | 26s |
| Go Test | ✅ pass | 33s |
| Lint | ✅ pass | 20s |
| GitGuardian Security Checks | ✅ pass | 1s |

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,34 @@
+.DS_Store
+
+# Ignore docker files.
+.dockerignore
+**/Dockerfile
+
+# Ignore git directory and files.
+.git
+.github
+.gitignore
+
+# Ignore editor directories and files.
+.idea
+.vscode
+
+# Ignore worktrees and documentation.
+.worktrees
+docs
+
+# Ignore node modules.
+node_modules
+**/node_modules
+
+# Ignore cache/output directories. These use **/ so they match at any depth,
+# matching how .gitignore treats them.
+**/bin
+**/build
+**/out
+**/tmp
+**/dist
+**/@mf-types
+
+# Ignore environment variables.
+**/*.env*
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,10 +13,6 @@
 .idea
 .vscode
 
-# Ignore worktrees and documentation.
-.worktrees
-docs
-
 # Ignore node modules.
 node_modules
 **/node_modules
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,23 +54,6 @@ pnpm build
 TW_ENV=production go run ./server/cmd/tw
 ```
 
-### Container image
-
-The repository-root `Dockerfile` builds the frontend and the server into a single image, so running Teacher Workspace needs only a container runtime instead of the Go, Node, and pnpm toolchains. A local build targets your own machine's architecture; CI builds `linux/arm64`:
-
-```bash
-docker build -t teacher-workspace .
-docker run --rm -p 3000:3000 teacher-workspace
-```
-
-Open <http://localhost:3000>. The image ships no `.env`, so every setting comes from `TW_*` variables and can be overridden at run time:
-
-```bash
-docker run --rm -p 8080:8080 -e TW_SERVER_PORT=8080 teacher-workspace
-```
-
-CI builds the image on every pull request. It is not pushed to any registry.
-
 ### Common commands
 
 Server:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,23 @@ pnpm build
 TW_ENV=production go run ./server/cmd/tw
 ```
 
+### Container image
+
+The repository-root `Dockerfile` builds the frontend and the server into a single image, so running Teacher Workspace needs only a container runtime instead of the Go, Node, and pnpm toolchains. A local build targets your own machine's architecture; CI builds `linux/arm64`:
+
+```bash
+docker build -t teacher-workspace .
+docker run --rm -p 3000:3000 teacher-workspace
+```
+
+Open <http://localhost:3000>. The image ships no `.env`, so every setting comes from `TW_*` variables and can be overridden at run time:
+
+```bash
+docker run --rm -p 8080:8080 -e TW_SERVER_PORT=8080 teacher-workspace
+```
+
+CI builds the image on every pull request. It is not pushed to any registry.
+
 ### Common commands
 
 Server:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,88 @@
+# syntax=docker/dockerfile:1
+
+# ----------------------------------------
+# Host build stage
+# ----------------------------------------
+FROM node:24-alpine3.23 AS host-build
+
+# Keep in sync with the pnpm/action-setup version in .github/workflows/ci.yml,
+# so the image and CI resolve dependencies identically.
+ARG PNPM_VERSION=11.11.0
+
+WORKDIR /src
+
+RUN npm install --global pnpm@${PNPM_VERSION}
+
+# Fetch all dependencies for better layer caching. Every workspace package
+# matched by pnpm-workspace.yaml needs its manifest copied here, otherwise the
+# frozen lockfile install fails on the missing importer.
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
+COPY apps/host/package.json apps/host/
+
+# --ignore-scripts skips the root prepare script, which installs the lefthook
+# git hooks and fails without a .git directory.
+RUN pnpm install --frozen-lockfile --ignore-scripts
+
+# Only apps/, so an edit under server/ leaves this stage cached.
+COPY apps/ apps/
+
+# Build the frontend into apps/host/dist.
+RUN pnpm build
+
+# ----------------------------------------
+# Server build stage
+# ----------------------------------------
+FROM golang:1.26.5-alpine3.23 AS server-build
+
+# BuildKit populates TARGETARCH from the platform being built. Hardcoding it
+# would emit a binary for the wrong architecture whenever the build host is not
+# arm64, which fails at run time rather than at build time.
+ARG TARGETARCH
+
+WORKDIR /src
+
+# Fetch all dependencies for better layer caching.
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY server/ server/
+
+# Build the binary. CGO_ENABLED=0 keeps it static so it does not depend on the
+# runtime stage's libc.
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
+    go build -trimpath -ldflags="-s -w" -o /src/tw ./server/cmd/tw
+
+# ----------------------------------------
+# Production stage
+# ----------------------------------------
+FROM alpine:3.23
+
+WORKDIR /app
+
+# No apk install step: alpine already ships ca-certificates-bundle, which is the
+# trusted root store a static Go binary reads, and adduser/addgroup are busybox
+# builtins. Keeping the stage offline makes the build reproducible.
+
+# 1. Create a new user named `tw`.
+# 2. Change the permission of `app` folder to user `tw`.
+# 3. Change the current user from `root` to `tw`.
+RUN addgroup -S tw && \
+        adduser -S tw -G tw && \
+        chown tw:tw /app
+
+USER tw
+
+COPY --from=server-build --chown=tw:tw /src/tw /app/tw
+COPY --from=host-build --chown=tw:tw /src/apps/host/dist /app/dist
+
+# The image ships no .env, so every setting comes from the environment.
+# TW_BUILD_DIR is absolute because config.Validate resolves it against the
+# working directory.
+ENV TW_ENV=production \
+    TW_BUILD_DIR=/app/dist \
+    TW_SERVER_PORT=3000
+
+EXPOSE 3000
+
+# Exec form, so the server is PID 1 and receives SIGTERM for graceful shutdown.
+CMD ["/app/tw"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,11 +46,8 @@ RUN go mod download
 COPY server/ server/
 
 # Build the binary. CGO_ENABLED=0 keeps it static so it does not depend on the
-# runtime stage's libc. No GOOS/GOARCH needed: the golang base image for this
-# stage is already Linux and already matches --platform, so go build defaults
-# to the right OS and architecture.
-RUN CGO_ENABLED=0 \
-    go build -trimpath -ldflags="-s -w" -o /src/tw ./server/cmd/tw
+# runtime stage's libc.
+RUN CGO_ENABLED=0 go build -trimpath -ldflags="-s -w" -o /src/tw ./server/cmd/tw
 
 # ----------------------------------------
 # Production stage
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,18 +13,21 @@ WORKDIR /src
 
 RUN npm install --global pnpm@${PNPM_VERSION}
 
-# Fetch all dependencies for better layer caching. Every workspace package
-# matched by pnpm-workspace.yaml needs its manifest copied here, otherwise the
-# frozen lockfile install fails on the missing importer.
-COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
-COPY apps/host/package.json apps/host/
+# Fetch all dependencies into the local store for better layer caching. This
+# reads only pnpm-lock.yaml, so it doesn't need every workspace package's
+# manifest copied in ahead of time.
+COPY pnpm-lock.yaml pnpm-workspace.yaml ./
+RUN pnpm fetch
 
+# Only package.json and apps/, so an edit under server/ leaves this stage
+# cached.
+COPY package.json ./
+COPY apps/ apps/
+
+# --offline links from the store fetched above instead of hitting the network.
 # --ignore-scripts skips the root prepare script, which installs the lefthook
 # git hooks and fails without a .git directory.
-RUN pnpm install --frozen-lockfile --ignore-scripts
-
-# Only apps/, so an edit under server/ leaves this stage cached.
-COPY apps/ apps/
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+# ----------------------------------------
+# Server build stage
+# ----------------------------------------
+FROM golang:1.26.5-alpine3.23 AS server-build
+
+# BuildKit populates TARGETARCH from the platform being built. Hardcoding it
+# would emit a binary for the wrong architecture whenever the build host is not
+# arm64, which fails at run time rather than at build time.
+ARG TARGETARCH
```
santosral, line 40: I think we don't need the GOARCH and it will default to the container's architecture.

1. Remove the `TARGETARCH` and `GOARCH`.
2. Add `RUN echo RESULT  GOARCH=$(go env GOARCH)"` before `go build`
3. Run  `docker build --no-cache --platform linux/arm64`. You can change the platform flag `linux/arm64` or `linux/amd64`.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+# ----------------------------------------
+FROM node:24-alpine3.23 AS host-build
+
+# Keep in sync with the pnpm/action-setup version in .github/workflows/ci.yml,
+# so the image and CI resolve dependencies identically.
+ARG PNPM_VERSION=11.11.0
+
+WORKDIR /src
+
+RUN npm install --global pnpm@${PNPM_VERSION}
```
santosral, line 14: I suggest to use pnpm base image (https://pnpm.io/docker).

```docker
FROM ghcr.io/pnpm/pnpm:11.11

WORKDIR /src
```

Then set this on `package.json` (https://pnpm.io/package_json#devenginesruntime, https://pnpm.io/package_json#devenginespackagemanager):

```json
  "devEngines": {
    "packageManager": {
      "name": "pnpm",
      "version": "^11.11.0",
      "onFail": "download"
    },
    "runtime": {
      "name": "node",
      "version": "^24.19.0",
      "onFail": "download"
    }
  }
```

The node version will be easier to manage and lessen the file changes.

Then on CI for node setup we can use [pnpm/setup@v2](https://github.com/pnpm/setup) which we also use the node and pnpm version from our `devEngines`

.dockerignore
```diff
@@ -0,0 +1,34 @@
...
+.github
+.gitignore
+
+# Ignore editor directories and files.
+.idea
+.vscode
+
+# Ignore worktrees and documentation.
+.worktrees
+docs
```
santosral, line 18: We don't have worktrees or docs for now.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+
+# Fetch all dependencies for better layer caching.
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY server/ server/
+
+# Build the binary. CGO_ENABLED=0 keeps it static so it does not depend on the
+# runtime stage's libc.
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=${TARGETARCH} \
```
santosral, line 52: I think we also don't need GOOS assignment here, same reason with GOARCH

Dockerfile
YimingIsCOLD: I think we can continue to use node image + the following way to install pnpm.

```dockerfile
# Install `pnpm`.
ENV PNPM_HOME="/pnpm"
ENV PATH="$PNPM_HOME:$PATH"
RUN mkdir $PNPM_HOME && \
        wget -qO- "https://github.com/pnpm/pnpm/releases/download/v10.33.4/pnpm-linuxstatic-arm64" > "$PNPM_HOME/pnpm" && \
        chmod +x $PNPM_HOME/pnpm && \
        ln -s $PNPM_HOME/pnpm /usr/local/bin/pnpm
```

CONTRIBUTING.md
```diff
@@ -54,6 +54,23 @@ pnpm build
 TW_ENV=production go run ./server/cmd/tw
 ```
 
+### Container image
```
YimingIsCOLD, line 57: I suggest we don't add this section now. Wait till we know the entire deployment flow then add.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+ARG PNPM_VERSION=11.11.0
+
+WORKDIR /src
+
+RUN npm install --global pnpm@${PNPM_VERSION}
+
+# Fetch all dependencies for better layer caching. Every workspace package
+# matched by pnpm-workspace.yaml needs its manifest copied here, otherwise the
+# frozen lockfile install fails on the missing importer.
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
```
YimingIsCOLD, line 19: Can consider if we would want to use `pnpm fetch`.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+WORKDIR /app
+
+# No apk install step: alpine already ships ca-certificates-bundle, which is the
+# trusted root store a static Go binary reads, and adduser/addgroup are busybox
+# builtins. Keeping the stage offline makes the build reproducible.
+
+# 1. Create a new user named `tw`.
+# 2. Change the permission of `app` folder to user `tw`.
+# 3. Change the current user from `root` to `tw`.
+RUN addgroup -S tw && \
```
YimingIsCOLD, line 69: Suggest we stick to user named `zero` as it is more generic.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+        chown tw:tw /app
+
+USER tw
+
+COPY --from=server-build --chown=tw:tw /src/tw /app/tw
+COPY --from=host-build --chown=tw:tw /src/apps/host/dist /app/dist
+
+# The image ships no .env, so every setting comes from the environment.
+# TW_BUILD_DIR is absolute because config.Validate resolves it against the
+# working directory.
```
YimingIsCOLD, line 80: No need to mention `.env` or `config.Validate` in this file.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+ENV TW_ENV=production \
+    TW_BUILD_DIR=/app/dist \
+    TW_SERVER_PORT=3000
+
+EXPOSE 3000
+
+# Exec form, so the server is PID 1 and receives SIGTERM for graceful shutdown.
```
YimingIsCOLD, line 87: Remove this comment.

Dockerfile
```diff
@@ -0,0 +1,88 @@
...
+# ----------------------------------------
+# Production stage
+# ----------------------------------------
+FROM alpine:3.23
+
+WORKDIR /app
+
+# No apk install step: alpine already ships ca-certificates-bundle, which is the
+# trusted root store a static Go binary reads, and adduser/addgroup are busybox
+# builtins. Keeping the stage offline makes the build reproducible.
```
YimingIsCOLD, line 64: Remove this comment.

Dockerfile
kellylimmm: Thanks both — tried both suggestions here, neither ends up working out for this repo, so I kept `npm install --global pnpm@${PNPM_VERSION}` as-is.

**Standalone binary (`YimingIsCOLD`):** the version needed to satisfy this repo's `engines.pnpm: ^11.0.0` (11.x) segfaults as a standalone binary on arm64 Alpine: confirmed across every 11.x release from `11.0.0` to the current `11.21.0`, checksums verified so it's not a bad download. The one version that doesn't crash (`10.33.4`, as suggested) gets rejected by our `engines.pnpm` field instead. No version clears both bars.

**`ghcr.io/pnpm/pnpm` base image (`santosral`):** pulls fine with the right pnpm version, but ships no Node.js and is Debian-based rather than Alpine (every other stage here is Alpine). Installing Node via pnpm's own `pnpm runtime set node` avoids needing `apt`, but was inconsistent in my testing: worked once via a plain container run, failed outright as part of an actual `docker build`. Given that flakiness, I don't want to depend on it for something that needs to build reliably in CI.

`npm install --global pnpm@11.11.0` doesn't have either problem: it satisfies our engine requirement, isn't affected by the native-binary bug, and keeps this stage on Alpine like the rest of the file. Happy to revisit either approach if circumstances change: e.g. the `devEngines` idea from the base-image suggestion is still worth doing independently, since it'd cut down version duplication between `package.json`, CI, and the Dockerfile.

Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+WORKDIR /src
+
+# Fetch all dependencies for better layer caching.
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY server/ server/
+
+# Build the binary. CGO_ENABLED=0 keeps it static so it does not depend on the
+# runtime stage's libc. No GOOS/GOARCH needed: the golang base image for this
```
YimingIsCOLD: I think no need to mention `GOOS/GOARCH`. Usually only those in used then we comment why need it.

Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+# stage is already Linux and already matches --platform, so go build defaults
+# to the right OS and architecture.
+RUN CGO_ENABLED=0 \
```
YimingIsCOLD: Can put them on the same line, no need `\`.

Dockerfile
kellylimmm: Hahahaha eh ya why is it included, removed!

Dockerfile
santosral: I'm good, I'll leave resolving this comment to @YimingIsCOLD for his suggestion.

kellylimmm: We can relook at this again, for now i'll leave it as `npm install --global pnpm@11.11.0`.

santosral (approved):

PR #74, opened 2026-08-07
feat(apps/mock-edupass): local OIDC provider with form_post and PKCE

### LINKED ISSUE #27: feat(`apps/mock-edupass`): local OIDC provider with `form_post` and PKCE
## User story

As a developer working on teacher-workspace, I want a lightweight local OIDC provider that supports `response_mode=form_post` and PKCE, so that I can build and test Edupass sign-in locally and in CI without Edupass onboarding, credentials, or MOE network access.

## Background

#12 makes teacher-workspace an OIDC relying party against Edupass (MOE's identity provider, backed by Microsoft Entra ID). Edupass requires onboarding, real credentials, and MOE network access, none of which exist on a developer machine or in CI, so the entire sign-in surface is untestable without a local stand-in.

Spike #16 (closed) chose **[`panva/node-oidc-provider`](https://github.com/panva/node-oidc-provider) v9**, an OpenID-certified library that covers the contract by configuration alone: `response_mode=form_post`, mandatory PKCE, a confidential client, `openid`-only with all profile claims packed into the ID token, plus discovery and JWKS endpoints. It eliminated the in-house `transformteamsg/mock-oidc-provider`, whose base (`oauth2-mock-server`) hard-codes `response_modes_supported: ['query']` and has no `form_post` path, and rejected Keycloak on weight. Its proof of concept verified the flow end to end. This issue builds the real provider from that proof of concept, hosted on **Express** and verified by its own test client, so the work does not wait on #12.

The library also owns its routes, including discovery and JWKS, and renders the `form_post` response itself: an auto-submitting HTML form whose action is the redirect URI, with a `<noscript>` fallback button. No endpoint or page templating is needed here beyond the `login` interaction. Two consequences for an Express host: body parsers such as `express.json()` must not be mounted ahead of the provider, which reads request bodies off the raw stream itself, and any Content-Security-Policy header must be applied before the provider handles the request, because the library appends a `sha256-` hash of its inline script to an existing `script-src` directive rather than using a nonce.

### Open questions (to resolve during grooming)

- **Package and directory name.** `apps/mock-edupass` is a placeholder chosen so this issue has a title and branch name. The final name is undecided.
- **How a developer selects the fake account at the `login` interaction.** Three candidates:
  - a page listing the hard-coded fake accounts, one click to pick one
  - auto-login as a single fake account, with no page at all
  - a standard email and password form

### Provider starts from a documented command

- **Given** a fresh checkout with dependencies installed
- **When** the developer runs the documented start command from the repository root
- **Then** the provider listens on its configured port and serves its discovery document, requiring no credentials and no network access beyond localhost

### Discovery and JWKS advertise the required capabilities

- **Given** the provider is running
- **When** a client fetches the OIDC discovery document
- **Then** `response_modes_supported` includes `form_post`, `code_challenge_methods_supported` includes `S256`, and the advertised `jwks_uri` serves the public key whose key ID matches the one used to sign ID tokens

### Authorization response reaches the client callback as a form post

- **Given** the provider is running with at least one fake account configured
- **When** a test client starts an authorization code flow requesting `scope=openid` with `response_mode=form_post` and an S256 PKCE challenge, and authentication completes as that fake account
- **Then** the provider returns an HTML document that submits `code` and `state` to the client's registered redirect URI as an HTTP POST body
- **And** no authorization response parameter appears anywhere in a URL

### ID token carries the fake account's claims

- **Given** a fake account configured with a subject identifier, an email address, and a name
- **When** the test client exchanges the authorization code using its PKCE verifier and client secret
- **Then** the response contains an ID token that verifies against the published JWKS and carries that account's `sub`, `email`, and `name`, even though only the `openid` scope was requested
- **And** a claim left unset on the fake account is absent from the ID token rather than present as null or an empty string

### Authorization request without PKCE is rejected

- **Given** the provider is running
- **When** an authorization request omits `code_challenge`
- **Then** the provider returns an OIDC error and issues no authorization code

### Token exchange with an invalid credential is rejected

- **Given** the provider has issued an authorization code
- **When** the token exchange presents a mismatched PKCE `code_verifier`, an incorrect client secret, or a code that has already been redeemed
- **Then** the provider returns an OIDC error response and issues no tokens

## Out of scope

- The outbound SCIM 2.0 provisioning client that pushes deactivation and reactivation. It was bundled into the earlier draft of this issue and is removed here; it will be filed separately if still needed.
- Edupass User Schema claim names. Only `sub`, `email`, and `name` are emitted. Mapping to real Edupass claim names is a later, configuration-only change and depends on the MOE-internal Interface Specifications for Relying Parties (v2.2.0).
- Container image and cloud development environment wiring (Dockerfile, compose service). Follow-up issue.
- The relying party side: the Go server's OIDC client, session creation, and the configuration that points it at this provider (#12).
- Any production or staging deployment. This provider is for local development and CI only.
- A consent screen. Whether Edupass prompts for consent at all is still to be researched, so no consent interaction is built here: the provider grants the requested `openid` scope without prompting.
- Security headers. This app sets no `Content-Security-Policy` and adds no `helmet` middleware, so the library's inline `form_post` script runs unimpeded. The ordering constraint described above applies whenever a policy is added later. Tracked in #63.
- Signing key management. The provider runs on `node-oidc-provider`'s bundled development key (`kid: keystore-CHANGE-ME`), which is stable across restarts and satisfies every criterion above, at the cost of a startup warning and a private key that is published in a public npm package. Supplying a generated key through the `jwks` option is tracked in #64.
- Storage. Fake accounts are hard-coded, and the library's default in-memory adapter is sufficient. No database, no configuration file, no environment variables for account data.
- Refresh tokens, RP-initiated logout / end-session, and the UserInfo endpoint beyond whatever the library provides by default.
- Registering the real relying party client in Edupass, SAML, real MOE credentials, and real teacher data.

## Design assets

No Figma assets. Sign-in flow, where the provider stands in for Edupass:

```mermaid
sequenceDiagram
    participant C as Test client
    participant B as Browser
    participant P as Local OIDC provider (this issue)

    C->>C: Generate state, nonce, PKCE verifier and S256 challenge
    C-->>B: Redirect to /authorize (scope=openid, response_mode=form_post, challenge)
    B->>P: GET /authorize
    P-->>B: Redirect to login interaction (selection mechanism is an open question)
    B->>P: Complete authentication as a fake account
    P-->>B: HTML page with auto-submitting form (action = redirect URI)
    B->>C: POST code + state to the callback
    C->>P: POST /token (code, code_verifier, client secret)
    P-->>C: ID token signed by the JWKS key, carrying sub / email / name
    C->>P: GET /jwks
    P-->>C: Public key, verifies the ID token signature
```

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

evtpano:
Close #27 

## 🚀 Summary

Add a local OIDC provider (`apps/mock-edupass`) that stands in for Edupass during development and CI testing. This lets developers work on authentication flows without real credentials.

## ✏️ Changes

- Scaffolded `apps/mock-edupass` package with Express + `oidc-provider`
- Implemented Authorization Code flow with PKCE (S256), `form_post` response mode, and `client_secret_post` token auth
- Added 3 hard-coded test accounts (`teacher-1`, `teacher-2`, `teacher-3`) with auto-login (no UI)
- Account selection via `?account=<id>` query param (defaults to `teacher-1`)
- Added test suite covering discovery, full OIDC flow, and error cases (8 tests)
- Added README documenting quick start, client configuration, and fake accounts

## 🧪 Test Plan

- `pnpm --filter @teacher-workspace/mock-edupass test` passes (8 tests)
- Verified discovery endpoint returns correct issuer, response modes, and code challenge methods
- Verified full authorization code flow with PKCE produces a valid signed JWT
- Verified error cases: missing PKCE rejected, wrong verifier rejected, wrong secret rejected, code reuse rejected

## Notes

**Auto-login (no UI):** Provider is dev/CI only. Auto-login keeps tests deterministic.

**Sample authorization URL:**

```
CODE_VERIFIER=$(openssl rand -base64 32 | tr -d '=' | tr '+/' '-_')
CODE_CHALLENGE=$(printf '%s' "$CODE_VERIFIER" | openssl dgst -sha256 -binary | base64 | tr -d '=' | tr '+/' '-_')

curl -v -L -b /tmp/cookies -c /tmp/cookies "http://localhost:9000/authorize?client_id=teacher-workspace&redirect_uri=http://localhost:3000/auth/callback&response_type=code&scope=openid&response_mode=form_post&code_challenge=${CODE_CHALLENGE}&code_challenge_method=S256&&account=teacher-2"
```

## Acceptance criteria coverage:

- **Discovery & JWKS advertise required capabilities** — `.well-known/openid-configuration` returns correct issuer, response modes, and code challenge methods; JWKS endpoint serves a valid signing key.
  - `discovery document has required fields`
  - `JWKS serves a valid public key`

- **Authorization response reaches callback as form post** — Full auth code flow with PKCE completes and delivers the code via HTML form POST to the redirect URI.
  - `completes authorization code flow with PKCE and form_post`

- **ID token carries fake account's claims** — Token exchange returns a signed JWT with correct `sub`, `email`, `name`; missing claims are omitted (not null).
  - `completes authorization code flow with PKCE and form_post`
  - `absent claim is not present in ID token (teacher-3)`

- **Authorization without PKCE is rejected** — Requests missing `code_challenge` are refused.
  - `rejects authorization without PKCE`

- **Token exchange with invalid credential is rejected** — Wrong verifier, wrong client secret, and reused authorization codes all fail.
  - `rejects token exchange with wrong code_verifier`
  - `rejects token exchange with wrong client_secret`
  - `rejects token exchange with reused code`


nwsgerald:
# Code Review — feat/fake-edupass (2026-08-12 08:41)

> **File:** `review/feat-fake-edupass/report-20260812084115.md`
> **Based on:** full branch diff since diverging from `main` (`71f9f1c`) — 12 files, +1061/−94
> **PR:** #74 · **Closes:** #27

## Summary

| Severity | Count |
|----------|-------|
| 🔴 Important    | 3 |
| 🟡 Nit          | 4 |
| 🟣 Pre-existing | 0 |

The provider itself is sound: PKCE, `form_post`, and the confidential client are configured against a certified library rather than hand-rolled, the ID token omits unset claims exactly as #27 requires, and the error-case tests assert specific OIDC error codes instead of just non-200s. All 8 tests pass locally and `tsc --noEmit` is clean. The risk is not in the OIDC logic but around it — the README documents an authorization route that returns 404, and nothing in CI runs the test suite or the type checker, so both the code and its documentation are unguarded from the moment this merges. Recommended next step: fix the README route (one line), add a CI job that runs `pnpm --filter @teacher-workspace/mock-edupass test` and `typecheck`, then merge; the nits can follow.

---

## Findings

### 🔴 Important

#### 1. README's only `account` example points at a route that returns 404

**File:** `apps/mock-edupass/README.md` · **Line:** 49 · **Triage:** To be fixed

```markdown
## How It Works

There is no login page or consent screen. Authentication and consent complete automatically:

- Defaults to `teacher-1` unless `?account=<id>` is passed to the authorize endpoint
- Example: `/auth?...&account=teacher-2` logs in as John Smith   <!-- ← /auth is not a route here -->
```

The provider overrides the authorization route in `src/provider.ts:30-32`:

```ts
routes: {
  authorization: '/authorize',
},
```

**Problem:** `/auth` is `oidc-provider`'s default authorization route, but this provider overrides it to `/authorize`. Verified against a running instance: `GET /auth?client_id=...` returns **404**, `GET /authorize?client_id=...` returns **303**, and the discovery document advertises `authorization_endpoint: http://localhost:9000/authorize`. This is the README's only example of the `account` parameter — the feature a developer is most likely to copy — and it fails outright. The PR description's sample curl uses `/authorize` correctly, so the defect is confined to the README.

**Suggestion:**
```markdown
- Example: `/authorize?...&account=teacher-2` logs in as John Smith
```

---

#### 2. The new test suite never runs in CI

**File:** `.github/workflows/ci.yml` · **Line:** 15 · **Triage:** To be fixed

```yaml
jobs:
  format:      # oxfmt
  lint:        # oxlint
  go-lint:     # golangci-lint
  go-test:     # go test -race ./...
  # ← no JS/TS test job
```

**Problem:** this PR adds the repository's first JS/TS test suite (8 tests), and the PR's test plan rests entirely on it, but no CI job executes it. `lefthook.yml` only runs `oxfmt` and `oxlint` on staged files at pre-commit. Nothing runs `pnpm test` on any pull request. The provider can break — a dependency bump, a config change, a route rename — and CI stays green. That suite is the sole evidence for five of the six acceptance criteria on #27, and it is unenforced the moment it merges.

**Suggestion:**
```yaml
  test:
    name: Test
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
      - name: Setup PNPM
        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
        with:
          version: 11.11.0
      - name: Setup Node.js
        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
        with:
          node-version: 24
          cache: pnpm
      - name: Install dependencies
        run: pnpm install --frozen-lockfile
      - name: Test
        run: pnpm --filter @teacher-workspace/mock-edupass test
```

> Add a root `"test": "pnpm -r test"` script if you'd rather the job stay generic as more packages gain tests.

---

#### 3. Missing automated test for: provider starts from a documented command

**File:** `apps/mock-edupass/test/provider.test.ts` · **Line:** 96 · **Triage:** To be fixed

```ts
before(async () => {
  const { app } = createApp(TEST_PORT);   // ← in-process; index.ts and config.ts never run
  server = app.listen(TEST_PORT);
  await new Promise<void>((resolve) => server.once('listening', resolve));
});
```

**Problem:** issue #27 lists six acceptance criteria; the PR's "Acceptance criteria coverage" section maps five and omits *"Provider starts from a documented command"*. Every test calls `createApp()` directly, so `src/index.ts`, `src/config.ts`, and the `MOCK_EDUPASS_PORT` environment variable are never exercised. Nothing verifies that the documented start command boots a server that serves discovery — and that untested seam is exactly where finding #1 lives: the docs and the config disagree, and no test could have caught it.

Verified manually during this review: `MOCK_EDUPASS_PORT=9911 node --experimental-strip-types src/index.ts` starts, `GET /health` returns 200, and discovery returns `authorization_endpoint: http://localhost:9911/authorize`. The criterion holds — it just isn't guarded.

**Suggestion:** spawn the documented entry point and assert on discovery, so the README's URLs are covered by the same test.

```ts
import { spawn } from 'node:child_process';

it('starts from the documented command and serves discovery', async () => {
  const proc = spawn('node', ['--experimental-strip-types', 'src/index.ts'], {
    env: { ...process.env, MOCK_EDUPASS_PORT: '9877' },
  });
  try {
    await waitForHealth('http://localhost:9877/health');
    const doc = await (await fetch('http://localhost:9877/.well-known/openid-configuration')).json();
    assert.equal(doc.authorization_endpoint, 'http://localhost:9877/authorize');
  } finally {
    proc.kill();
  }
});
```

---

### 🟡 Nit

#### 4. An unknown `account` id silently authenticates as `teacher-1`

**File:** `apps/mock-edupass/src/interaction.ts` · **Line:** 14 · **Triage:** To be fixed

```ts
if (details.prompt.name === 'login') {
  const requestedAccount = details.params.account as string | undefined;
  const account = accounts.find((a) => a.sub === requestedAccount) ?? accounts[0];   // ←

  await provider.interactionFinished(
    req,
    res,
    { login: { accountId: account.sub } },
    { mergeWithLastSubmission: false },
  );
  return;
}
```

**Problem:** `?account=teacher-99` does not error — it authenticates as `teacher-1`. Verified end to end: the flow returns an authorization code and the ID token carries `sub=teacher-1`, `email=jane.doe@example.com`, `name=Jane Doe`. The `??` fallback is correct for an *absent* parameter but wrong for an *invalid* one. In a downstream relying-party test, a typo'd account id produces a passing flow with the wrong claims, surfacing later as a confusing assertion mismatch — if at all. This matters more than usual here, since deterministic account selection is the package's entire purpose.

**Suggestion:**
```ts
const requestedAccount = details.params.account as string | undefined;
const account = requestedAccount
  ? accounts.find((a) => a.sub === requestedAccount)
  : accounts[0];

if (!account) {
  throw new Error(
    `Unknown account '${requestedAccount}'. Available: ${accounts.map((a) => a.sub).join(', ')}`,
  );
}
```

> Express 5 forwards a rejected async handler to the error handler, so this surfaces as a 500 carrying the message rather than a silent wrong login.

---

#### 5. An unhandled prompt name sends no response and logs nothing

**File:** `apps/mock-edupass/src/interaction.ts` · **Line:** 38 · **Triage:** To be fixed

```ts
router.get('/interaction/:uid', async (req, res) => {
  const details = await provider.interactionDetails(req, res);

  if (details.prompt.name === 'login') { /* ... */ return; }

  if (details.prompt.name === 'consent') {
    /* ... */
    await provider.interactionFinished(req, res, { consent: { grantId } }, { mergeWithLastSubmission: true });
  }
  // ← falls through: no response written, nothing logged
});
```

**Problem:** if `prompt.name` is neither value, the handler returns without touching `res`, and the request hangs until the client or `server.requestTimeout` gives up — with nothing logged to explain it. Checked against `oidc-provider@9.11.1`: its default interaction policy defines exactly two prompts (`lib/helpers/interaction_policy/prompts/{login,consent}.js`), so this is unreachable today. It becomes reachable as soon as someone sets a custom `interactions.policy` or a major upgrade introduces a prompt.

**Suggestion:**
```ts
    await provider.interactionFinished(req, res, { consent: { grantId } }, { mergeWithLastSubmission: true });
    return;
  }

  throw new Error(`Unhandled interaction prompt: ${details.prompt.name}`);
});
```

---

#### 6. Hardcoded test port turns a port collision into an infinite hang

**File:** `apps/mock-edupass/test/provider.test.ts` · **Line:** 13 · **Triage:** To be fixed

```ts
const TEST_PORT = 9876;                                                     // ←
// ...
before(async () => {
  const { app } = createApp(TEST_PORT);
  server = app.listen(TEST_PORT);
  await new Promise<void>((resolve) => server.once('listening', resolve));  // ← never resolves
});
```

**Problem:** verified by occupying port 9876 and running the suite — `'listening'` never fires, and since node:test applies no default timeout, `pnpm test` hangs indefinitely. Forcing `--test-timeout=20000` reports `'test did not finish before its parent and was cancelled'`, which never mentions `EADDRINUSE`. On a developer machine already running the provider, or on a shared CI runner, this is a long unexplained stall instead of a one-line error.

The root cause is in `src/app.ts:6-7` — `createApp(port)` derives the issuer from a port it does not own, so the port must be known before `listen()` and an ephemeral port is impossible:

```ts
export function createApp(port: number) {
  const provider = createProvider(port);   // issuer = http://localhost:${port}
```

**Suggestion — Option A** (three lines, keeps the port hardcoded, only improves the error):
```ts
await new Promise<void>((resolve, reject) => {
  server.once('listening', resolve);
  server.once('error', reject);   // ← EADDRINUSE surfaces immediately
});
```

**Suggestion — Option B** (removes the coupling, immune to collisions):
```ts
export function createApp(issuer: string) { /* ... */ }

// index.ts
createApp(`http://localhost:${config.port}`);

// test
const server = app.listen(0);
const { port } = server.address() as AddressInfo;
const BASE_URL = `http://localhost:${port}`;
```

> B costs a two-line signature change across the only two callers and makes the suite collision-proof; A only improves the message.

---

#### 7. `tsconfig.json` and the `typescript` dependency are never executed

**File:** `apps/mock-edupass/package.json` · **Line:** 6 · **Triage:** To be fixed

```json
"scripts": {
  "start": "node --experimental-strip-types src/index.ts",
  "test": "node --experimental-strip-types --test test/**/*.test.ts"
},                                          // ← no typecheck script
"devDependencies": {
  "typescript": "^6.0.3"                    // ← never invoked
}
```

**Problem:** the package ships a `tsconfig.json` with `strict`, `noUnusedLocals`, and `noUnusedParameters`, but both scripts run under `--experimental-strip-types`, which strips type annotations *without checking them*. No `typecheck` script exists and no CI job runs `tsc`. `pnpm exec tsc --noEmit` passes cleanly today, so there is nothing to fix in the code — the gap is that those strictness settings guard nothing going forward.

**Suggestion:**
```json
"scripts": {
  "start": "node --experimental-strip-types src/index.ts",
  "test": "node --experimental-strip-types --test test/**/*.test.ts",
  "typecheck": "tsc --noEmit"
}
```

> Wire it into the CI job from finding #2.

---

## Reviewer To-Do

- Manually test: provider starts from the documented command and serves discovery on the configured port — *performed during this review, passes*
- Manually test: no authorization response parameter appears anywhere in a URL — the `form_post` test asserts the form action and body but never asserts the absence of `code`/`state` from a redirect URL

---

## What Looks Good

- **`findAccount` builds claims conditionally rather than emitting nulls** (`src/provider.ts:53-58`) — satisfies #27's "absent, not null or empty" criterion exactly, and the test asserts it with `Object.hasOwn` rather than a truthiness check that an empty string would pass
- **No body parser mounted ahead of `provider.callback()`** (`src/app.ts:12-17`) — the issue called this constraint out explicitly and it is honored; `/health` is registered with `app.get` so it cannot intercept the provider's POST routes
- **Error cases assert specific OIDC error codes** (`test/provider.test.ts:245-278`) — wrong verifier, wrong client secret, and code reuse each check for `invalid_grant` / `invalid_client` rather than a bare non-200, which is what makes the PKCE and confidential-client configuration trustworthy
- **PKCE enforced through configuration, not hand-rolled validation** (`src/provider.ts:26-28`) — `required: () => true` delegates to the certified library instead of re-implementing the check

nwsgerald:
the above is slop produced by review skill take it with a handful of salt

evtpano:
## Addressed Review Comments

**Review Point 1 — README's authorize route example (Important)**
Fixed `/auth?...&account=teacher-2` → `/authorize?...&account=teacher-2` in README.md.

**Review Point 4 — Unknown account silently authenticates as teacher-1 (Nit)**
Invalid account ids now throw an error listing available accounts instead of falling back silently. Verified end-to-end: `?account=teacher-99` returns a 500 with `Unknown account 'teacher-99'. Available: teacher-1, teacher-2, teacher-3`.

**Review Point 5 — Unhandled prompt name sends no response (Nit)**
Added explicit `return` after consent handling and a `throw` for unrecognized prompt names, so the request fails loudly instead of hanging.

**Review Point 6 — Hardcoded test port causes infinite hang (Nit)**
Added `server.once('error', reject)` to the test setup so a port collision surfaces immediately as an `EADDRINUSE` error instead of hanging indefinitely.

**Review Point 7 — typecheck script missing (Nit)**
Added `"typecheck": "tsc --noEmit"` to package.json scripts.

## Not addressed (for now)

**Review Point 2 — CI test job**
Deferred. Currently we only have a single `ci.yml` and need to discuss the monorepo CI strategy first (running jobs only for packages with changes). TBD

## Not addressing

**Review Point 3 — Test for documented start command**
The OIDC logic is covered by the existing in-process tests. The doc/code mismatch this finding would guard against is fixed directly by Review Point 1

nwsgerald (dismissed):
lgtm

evtpano:
Rebased and downgraded existing typescript from "^6.0.3" to "^5.9.3" with reference to #91

apps/mock-edupass/README.md
```diff
@@ -0,0 +1,61 @@
+# mock-edupass
+
+A local OIDC provider that stands in for Edupass during development and CI testing. It implements the OpenID Connect Authorization Code flow with PKCE, auto-login (no UI), and hard-coded test accounts — so you can develop and test authentication without real credentials or MOE network access.
```
nwsgerald: think contributing say dont use em-dash

apps/mock-edupass/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@teacher-workspace/mock-edupass",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "start": "node --experimental-strip-types src/index.ts",
+    "test": "node --experimental-strip-types --test test/**/*.test.ts",
```
nwsgerald: i think kelly already pinned 24.0.0 at package.json, is this still necessary

apps/mock-edupass/test/provider.test.ts
```diff
@@ -0,0 +1,283 @@
...
+
+function decodeJwtPayload(jwt: string): Record<string, unknown> {
+  const payload = jwt.split('.')[1];
+  return JSON.parse(Buffer.from(payload, 'base64url').toString());
+}
+
+function decodeJwtHeader(jwt: string): Record<string, unknown> {
+  const header = jwt.split('.')[0];
+  return JSON.parse(Buffer.from(header, 'base64url').toString());
+}
```
nwsgerald: identical except for the index; possible to merge?

nwsgerald (changes requested):

apps/mock-edupass/README.md
evtpano: Removed em-dash from README.md

apps/mock-edupass/package.json
evtpano: removed --experimental-strip-types since node 24 default support

apps/mock-edupass/test/provider.test.ts
evtpano: Merged the logic into a shared decodeJwtPart but kept the named wrappers so call sites read as decodeJwtPayload(token) / decodeJwtHeader(token) rather than requiring readers to remember which index maps to which JWT segment.

apps/mock-edupass/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "start": "node src/index.ts",
+    "test": "node --test test/**/*.test.ts",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "express": "^5",
```
YimingIsCOLD: Would prefer to specific the minor and patch version too.

apps/mock-edupass/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { createApp } from './app.ts';
+import { config } from './config.ts';
+
+const { app } = createApp(config.port);
+
+app.listen(config.port, () => {
+  // eslint-disable-next-line no-console
```
YimingIsCOLD:
```suggestion
  // oxlint-disable-next-line no-console
```

evtpano: replaced with `oxlint`

apps/mock-edupass/package.json
evtpano: added the minor and patch version for `express` & `oidc-provider`

evtpano:
With reference to [Slack message](https://gvt-external.slack.com/archives/C0BMBF196R1/p1786938348568359), ok to merge since this will likely remain a small mock app. Merged `accounts.ts` into `provider.ts` and `interaction.ts` into `app.ts`. 

Kept `index.ts` separate from `app.ts` because the test imports `createApp` from `app.ts`. Merging them would require a main-module guard to prevent the server auto-starting on import during test runs.

YimingIsCOLD (approved):

nwsgerald (approved):

PR #81, opened 2026-08-13
chore(deps): update vulnerable transitive dependencies

kellylimmm:
## Summary

This PR is reset to track `main` and will stay focused solely on the `undici` vulnerability (plus `adm-zip`, which is fixed by the same upstream release).

The fix requires bumping `@module-federation/dts-plugin` (and the rest of the lockstep `@module-federation/*` family) to `2.8.2`. That release was published 2026-08-06T11:24 UTC, which is inside this workspace's 7-day `minimumReleaseAge` cooldown (`pnpm-workspace.yaml`). Per team decision, this PR stays open and empty until the cooldown clears (~2026-08-13T11:24 UTC), at which point the `@module-federation/*` bump will be pushed here to resolve the remaining 12 `undici` alerts and 1 `adm-zip` alert.

The other 25 of the original 38 Dependabot alerts have been split into smaller PRs merging into this branch:
- #82 — hono, @hono/node-server, body-parser, postcss, nanoid, js-yaml, brace-expansion, ip-address, fast-uri (via shadcn's transitive deps)
- #83 — react-router (RSC mode CSRF bypass)

## Test plan

- [ ] Waiting on cooldown before applying the @module-federation/* bump
- [ ] `pnpm audit`, `pnpm lint`, `pnpm build` to be re-verified once the fix is applied

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,30 @@
+.DS_Store
+
+# Ignore docker files.
+.dockerignore
+**/Dockerfile
+
+# Ignore git directory and files.
+.git
+.github
+.gitignore
+
+# Ignore editor directories and files.
+.idea
+.vscode
+
+# Ignore node modules.
+node_modules
+**/node_modules
+
+# Ignore cache/output directories. These use **/ so they match at any depth,
+# matching how .gitignore treats them.
+**/bin
+**/build
+**/out
+**/tmp
+**/dist
+**/@mf-types
+
+# Ignore environment variables.
+**/*.env*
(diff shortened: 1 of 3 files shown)


YimingIsCOLD:
Too many dependencies upgrade in this PR. I would suggest resetting this branch and splitting the upgrades into several smaller PRs that merge into this branch. Each PR could upgrade an individual dependency or a group of related dependencies where it makes sense.

This PR can remain focused on fixing the undici vulnerability and be the only PR that merges into main.

kellylimmm:
This was the original PR meant to merge the staging branch into `main`. It got auto-closed by GitHub as a side effect of resetting the branch, once the branch had zero commits ahead of `main`, GitHub closes the PR automatically since there's nothing left to merge. It's not reopenable in that state, so once the staging branch is ready to ship, we'll open a fresh PR into `main` instead of reusing #81's number.

PR #98, opened 2026-08-17
chore(`server/httputil`): add test for `RenderPlain` and `RenderJSON`

### LINKED ISSUE #99: chore(server/httputil): add test for RenderPlain and RenderJSON
## User story

As an Engineer, I want the plain text and JSON response renderers covered by tests, so that a regression in their headers, status codes, bodies, or failure logging is caught before reaching production.

## Background

`server/internal/httputil` holds the two renderers to write a response: one for plain text and one for JSON. Both set `Content-Type` and `X-Content-Type-Options: nosniff`, write the status code, write the body, and log when the body cannot be written.

The package currently has no tests. Changed header value, or a write failure would affect every response without failing a test.

### Plain text renderer sets the headers and status code

- **Given** a recorded HTTP response
- **When** the plain text renderer is called with a status
- **Then** the response has `Content-Type: text/plain; charset=UTF-8`, has `X-Content-Type-Options: nosniff`, and returns the status code that was passed in

### Plain text renderer writes the status text as the body

- **Given** a recorded HTTP response
- **When** the plain text renderer is called with a status
- **Then** the body is that reason phrase, the body is empty for the status that has none, and nothing is logged

### Plain text renderer logs a failed body write

- **Given** a logger that writes to a buffer
- **When** the plain text renderer is called on a response that cannot hold a body
- **Then** one `ERROR` line is logged, and it names the plain text renderer and includes the error

### JSON renderer sets the headers and status code

- **Given** a recorded HTTP response
- **When** the JSON renderer is called with a status code, but no value.
- **Then** the response has `Content-Type: application/json; charset=UTF-8`, has `X-Content-Type-Options: nosniff`, status code that was passed in, and null body.

### JSON renderer writes the value as JSON

- **Given** a recorded HTTP response
- **When** the JSON renderer is called with a value
- **Then** the body is that value as JSON

### JSON renderer logs a failed encode

- **Given** a logger that writes to a buffer
- **When** the JSON renderer is called with a value that cannot be turned into JSON
- **Then** one `ERROR` line is logged, and it names the JSON renderer and includes the error

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

santosral:
Closes #99 

## 🚀 Summary

Added missing tests from `httputil` renderers functions.

## ✏️ Changes

- **`internal/httputil`**: adds `httputil_test.go` as an external test package (`httputil_test`), with ten subtests across `TestRenderPlain` and `TestRenderJSON`. Each function is covered for the `Content-Type` header, the `X-Content-Type-Options` header, the status code, the response body, and the write-failure log line.

## 🧪 Test Plan

- `go test ./server/internal/httputil/` passes
- `go vet ./server/internal/httputil/` clean
- `gofmt -l server/` no output

### DIFF
--- a/server/internal/httputil/httputil_test.go
+++ b/server/internal/httputil/httputil_test.go
@@ -1,0 +1,139 @@
+package httputil_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/String-sg/teacher-workspace/server/internal/httputil"
+	"github.com/String-sg/teacher-workspace/server/pkg/require"
+)
+
+func TestRenderPlain(t *testing.T) {
+	t.Run("sets Content-Type header", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+		res := rec.Result()
+
+		require.Equal(t, "text/plain; charset=UTF-8", res.Header.Get("Content-Type"))
+	})
+
+	t.Run("sets X-Content-Type-Options header", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+		res := rec.Result()
+
+		require.Equal(t, "nosniff", res.Header.Get("X-Content-Type-Options"))
+	})
+
+	t.Run("responds with status code from given status code", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+		res := rec.Result()
+
+		require.Equal(t, 500, res.StatusCode)
+	})
+
+	t.Run("responds with text from given status code", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+
+		require.Equal(t, "Internal Server Error", rec.Body.String())
+	})
+
+	t.Run("logs error message", func(t *testing.T) {
+		var logs bytes.Buffer
+		logger := slog.New(slog.NewJSONHandler(&logs, nil))
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusNoContent)
+
+		type logLine struct {
+			Level    string `json:"level"`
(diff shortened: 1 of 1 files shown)


server/internal/httputil/httputil_test.go
```diff
@@ -0,0 +1,139 @@
...
+
+func TestRenderPlain(t *testing.T) {
+	t.Run("sets Content-Type header", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+		res := rec.Result()
+
+		require.Equal(t, "text/plain; charset=UTF-8", res.Header.Get("Content-Type"))
```
YimingIsCOLD: Don't use `require`. We are deprecating it.

server/internal/httputil/httputil_test.go
```diff
@@ -0,0 +1,139 @@
...
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+		res := rec.Result()
+
+		require.Equal(t, "nosniff", res.Header.Get("X-Content-Type-Options"))
+	})
+
+	t.Run("responds with status code from given status code", func(t *testing.T) {
```
YimingIsCOLD:
```suggestion
	t.Run("responds with the expected status code", func(t *testing.T) {
```

I think we should avoid using `given` in test descriptions. Test names generally read more naturally when they describe the expected behaviour rather than the setup.

server/internal/httputil/httputil_test.go
```diff
@@ -0,0 +1,139 @@
...
+	t.Run("responds with text from given status code", func(t *testing.T) {
+		logger := slog.New(slog.DiscardHandler)
+		rec := httptest.NewRecorder()
+
+		httputil.RenderPlain(rec, logger, http.StatusInternalServerError)
+
+		require.Equal(t, "Internal Server Error", rec.Body.String())
+	})
+
+	t.Run("logs error message", func(t *testing.T) {
```
YimingIsCOLD: Need more context. `logs error message when ...`

server/internal/httputil/httputil_test.go
santosral: Removed the assertion helpers

server/internal/httputil/httputil_test.go
santosral: Updated the test descriptions for readability.

server/internal/httputil/httputil_test.go
santosral: Add the answers to "when" for clarity.

YimingIsCOLD (approved):

PR #100, opened 2026-08-17
chore(`apps/host/deps`): bump `postcss` to 8.5.26

### LINKED ISSUE #101: chore(`apps/host/deps`): bump `postcss` to 8.5.26
## User story

As an Engineer, I want `nanoid` on a patched version to fix the vulnerability.

## Background

`pnpm audit` reports two high severity advisories for `nanoid`. Both let a generator loop forever, which hangs the process.

- [#18](https://github.com/String-dxd/teacher-workspace/security/dependabot/18)
- [#39](https://github.com/String-dxd/teacher-workspace/security/dependabot/39) 
- [#40](https://github.com/String-dxd/teacher-workspace/security/dependabot/40)
- [#42](https://github.com/String-dxd/teacher-workspace/security/dependabot/42)

`nanoid` is a transitive dependency. `postcss` pulls it in. `postcss` is pulled in by `shadcn` and by `webpack`.

Updating `postcss` updates `nanoid` too.

Updating `nanoid` on its own does not hold. `postcss@8.5.15` asks for `nanoid: ^3.3.12`, so `nanoid` can drop back to 3.3.12 on the next install. `postcss@8.5.26` asks for `nanoid: ^3.3.17`, so 3.3.12 is no longer allowed.

Updating `postcss` also fixes two `postcss` advisories.

### nanoid resolves to a patched version

- **Given** a fresh clone of the repo
- **When** a developer runs `pnpm install` and then `pnpm why nanoid`
- **Then** every `nanoid` entry in the tree is 3.3.18 or higher

### Audit reports no nanoid advisory

- **Given** `postcss` is updated and the lockfile is committed
- **When** a developer runs `pnpm audit`
- **Then** no `nanoid` advisory is listed.

### No override is needed

- **Given** `postcss@8.5.26` asks for `nanoid: ^3.3.17`
- **When** pnpm resolves the dependency tree
- **Then** `nanoid` is patched with no `overrides` entry in `pnpm-workspace.yaml`

### Host app still builds and runs

- **Given** `postcss` is updated
- **When** a developer runs `pnpm build` and then `pnpm preview`
- **Then** the build succeeds and the preview serves the host app with no new errors

## Design assets

N/A. This is a dependency change with no UI.

---

## Technical context

_Pending grooming._

## Data model

_Pending grooming._

## API contract

_Pending grooming._

## Error contract

_Pending grooming._

## Additional test scenarios

_Pending grooming._

## Hard constraints

_Pending grooming._

---

_🤖 Generated with create-issue_

santosral:
Closes #101

## 🚀 Summary

This PR updates `postcss` to a patched version to fix the vulnerabilities from `postcss` and `nanoid` because nanoid is a transitive dependency of it.

I've tried to update the `nanoid` using `pnpm update `or `pnpm audit --fix update -i `which doesn't do anything.

## ✏️ Changes

- Updated `postcss` 8.5.15 to 8.5.26, `nanoid` 3.3.12 to 3.3.18.
- Updated `pnpm-lock.yaml`.

No `overrides` entry is needed. postcss dependency of `^nanoid@3.3.17` which means it accepts minor and patch versions starting from the given version.

## 🧪 Test Plan

Build
```zsh 
pnpm build
```
Preview prod build
```zsh 
pnpm preview
```

Check if nanoid vulnerability is still there.
```zsh 
pnpm audit
```

### Video

https://github.com/user-attachments/assets/7d380d03-f0ac-43de-9438-33d3e38d2fcb

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,6 +8,3 @@ allowBuilds:
   lefthook: false
 
 minimumReleaseAge: 10080
-
-overrides:
-  nanoid@<3.3.18: ^3.3.18
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -8,3 +8,6 @@ allowBuilds:
   lefthook: false
 
 minimumReleaseAge: 10080
+
+overrides:
+  nanoid@<3.3.18: ^3.3.18
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,9 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  nanoid@<3.3.18: ^3.3.18
+
 importers:
 
   .:
@@ -28,7 +31,7 @@ importers:
         version: 5.2.8
       '@module-federation/enhanced':
         specifier: ^2.5.1
-        version: 2.5.1(@rspack/core@2.0.8(@module-federation/runtime-tools@2.5.1(node-fetch@2.7.0(encoding@0.1.13)))(@swc/helpers@0.5.23))(node-fetch@2.7.0(encoding@0.1.13))(typescript@5.9.3)(webpack@5.107.2(lightningcss@1.32.0)(postcss@8.5.15))
+        version: 2.5.1(@rspack/core@2.0.8(@module-federation/runtime-tools@2.5.1(node-fetch@2.7.0(encoding@0.1.13)))(@swc/helpers@0.5.23))(node-fetch@2.7.0(encoding@0.1.13))(typescript@5.9.3)(webpack@5.107.2)
       class-variance-authority:
         specifier: ^0.7.1
         version: 0.7.1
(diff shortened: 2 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -8,3 +8,6 @@ allowBuilds:
   lefthook: false
 
 minimumReleaseAge: 10080
+
+overrides:
```
YimingIsCOLD: Remove `overrides` once done upgrading. We will only keep it in some scenarios but not this.

YimingIsCOLD (commented):
We should upgrade `postcss` instead of `nanoid`. The latest version of `postcss` uses `nanoid: ^3.3.18`. Not sure what you have done, there is a lot of `supports-color` diff which should not have happen.

YimingIsCOLD:
> We should upgrade `postcss` instead of `nanoid`. The latest version of `postcss` uses `nanoid: ^3.3.18`. Not sure what you have done, there is a lot of `supports-color` diff which should not have happen.

I am wrong about latest version of `postcss` uses `nanoid: ^3.3.18`. The latest version of `postcss` uses `nanoid: ^3.3.17` which fixes one vulnerability but we need `nanoid` to be `3.3.18` to fix another vulnerability.

pnpm-workspace.yaml
santosral: Done! I've encountered some issues with `pnpm update nanoid` and `pnpm audit --fix update -i`where it doesn't work. I've tried `pnpm update postcss` instead where it updates the `nanoid` too.

kellylimmm (approved):

